PR #7, opened 2017-08-28
fixing various issues with dbench

carlos22:
- Added missing shebang
- Fixed malformed multi line variables
- Removed unnecessary su (which is not working in this way anyway)
- Added missing error handling (unsupported options as well as missing argument)
- Simplified the whole script (without the removal of features, hopefully?)

Note: It might be better to change -c to the default, and add a -f (or similar) to open the inside-docker shell (current default) (so dbench would be the same as bench).


vishalseshagiri:
Thanks

PR #14, opened 2017-10-08
[Cant merge] Fixed mariadb ports mapping

africlouds:
Port mapping on mariadb container should be 3307:3306 instead of 3307:3307 because internally mariadb listens to port 3306

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ services:
     volumes:
       - ./conf/mariadb-conf.d:/etc/mysql/conf.d
     ports:
-      - "3307:3307" #mariadb-port
+      - "3307:3306" #mariadb-port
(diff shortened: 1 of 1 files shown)


vishalseshagiri:
@africlouds thanks for the PR, well the mariadb container by default exposes its port 3306 to all other docker containers so this port mapping should'nt technically affect the working of your container. Having said that, you are right the port mapping was a glitch from my end during my last push.

africlouds:
Yeah I see. It only affects if you want to manage the DB with an external mysql client, such as navicat.

vishalseshagiri:
@africlouds  would really appreciate it if you could send this and all your PRs henceforth to the develop branch and not the master branch. Thanks and sorry for the trouble.

PR #29, opened 2018-01-18
Use patched qt

shanavas786:
Without patched qt, exporting to pdf fails.


Thunderbottom:
closing since wkhtmltopdf now works fine without requiring to be patched.

PR #30, opened 2018-01-20
Updates for dbench and .gitignore

sunhoww:
The rationale for using `cp` instead of `mv` for _Procfile_ and _common_site_config.json_ is to keep the repo clean, while still being able to have _bench_ and _frappe_ stuffs in the _frappe_bench_ folder.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+/data_dir
+/frappe-bench
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /frappe-bench/**/*
 !/frappe-bench/Procfile_docker
 !/frappe-bench/sites/common_site_config_docker.json
+/data_dir
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+/frappe-bench/**/*
+!/frappe-bench/Procfile_docker
+!/frappe-bench/sites/common_site_config_docker.json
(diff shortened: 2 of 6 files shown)


vishalseshagiri:
Didn't get your rationalization, with mv the files Procfile_docker and common_site_config_docker.json get renamed and are no longer parts of the repo which makes the repo look as desired. By doing a cp you'll have both Procfile and Procfile_docker in the directories which is not only a redundancy but also makes the contents of the directory look different from a non docker setup. I may have understood you wrong please elaborate what you meant by keeping the repo clean.

sunhoww:
What I meant by "clean" was to keep a local repo an exact clone of the upstream one.

I've been using **frappe_docker** for working on frappe apps. It works perfect for me. I create a bunch of sites and apps under _~/frappe_bench_. But when there are changes upstream, I could **not** just `pull`. The files under _frappe_bench_ locally would conflict.

What I would have like is to have left the files from this repo untouched. Having _frappe_bench_ outside of version control would solve this. I could just `pull` whenever there are changes upstream. This would mean leaving _Procfile_docker_ and _common_site_config_docker.json_, under _frappe_bench_, intact.

If we are to have the contents of _frappe_bench_ the same as that of a non docker setup, we could keep them outside of this folder altogether and `cp` into _frappe_bench_ during `dbench init`.

I hope I am not confusing things further.

I have been using this setup since November last. Works like a charm for me. Thanks.

PR #37, opened 2018-05-04
Updated README.md to clarify initial admin password

reilly3000:
it took me a long time to way to login after initially installing. I eventually found out the default by looking into closed issues on github. I just added a line to the readme to make it clear for future users to get up and running faster.


codingCoffee:
Thanks for the PR @reilly3000

PR #45, opened 2018-08-17
Specific mariadb version to 10.2

sopanawit:
Latest mariadb version is 10.3 which is not support required configuration for frappe-bench below.
```
innodb-file-format=barracuda
innodb-file-per-table=1
innodb-large-prefix=1
```


Thunderbottom:
this has already been fixed, thank you though.

PR #53, opened 2018-10-05
Tool to replace dbench

chabad360:
This PR adds the `bench` script, which aims for near parity with frappe bench, it makes setting up far easier, because instead of running `./dbench -c "new-site site1.local"` you only white what you would if bench was installed on the local system (i.e. `./bench new-site site1.local`), it also adds other features as well.


Thunderbottom:
rename the file to something that is not `bench`? `bench-docker` would be good I guess.

chabad360:
Just curious, Why?

Thunderbottom:
Just so to not confuse people with the actual bench command. Also, is this really necessary? Why not just append the changes to `dbench`

chabad360:
No issue doing that, I just didn't want to get "influenced" by the existing code, so I stared again, but I will do this.

chabad360:
Ready for merge

Issue #58, opened 2018-10-24
Code Sprint 2018 : Docker attempt


revant:
- [x] Switch to Alpine:latest
- [x] Manage redis-server dependency for bench setup in docker
- [x] <strike>Try docker-compose without links</strike>
- [x] Dockerfile cleanup
- [x] use custom build wkhtmltopdf (https://github.com/frappe/frappe_docker/issues/35) (https://github.com/frappe/frappe_docker/issues/33)
- [x] use distro's nodejs instead of nodesource.com
- [x] cleanup dbench

revant:
Closing Issue, continue discussion on PR

PR #59, opened 2018-10-25
Install Frappe using alpine image

revant:
Conf Sprint 2018

#58 

This Dockerfile and docker-compose for frappe:alpine is complete.
It has exposed ports and volume /home/frappe/frappe-bench/sites for accessing 

For Continuous Delivery, following steps need to be done

 - the Dockerfile and docker-entrypoint.sh need to be part of root frappe repo
 - instead of cloning frappe into bench init, frappe needs to be copied
 - in case frappe is copied instead of `bench init` then the config and installation steps will be manually done in Dockerfile

### DIFF
--- a/Dockerfile-frappe-alpine
+++ b/Dockerfile-frappe-alpine
@@ -52,7 +52,11 @@ RUN cd /home/frappe \
 
 WORKDIR /home/frappe/frappe-bench
 
+VOLUME /home/frappe/frappe-bench/sites
+
 EXPOSE 8000
+EXPOSE 9000
+EXPOSE 6787
 
 ENTRYPOINT ["docker-entrypoint.sh"]
 CMD ["start"]
--- a/Dockerfile-frappe-alpine
+++ b/Dockerfile-frappe-alpine
@@ -1,0 +1,58 @@
+# Frappe Bench
+FROM alpine:latest
+MAINTAINER developers@frappe.io
+
+USER root
+ENV LANG C.UTF-8
+
+COPY docker-entrypoint.sh usr/local/bin/docker-entrypoint.sh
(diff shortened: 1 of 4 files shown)


achillesrasquinha:
What's the resultant size?

Thunderbottom:
do we really need a custom build for wkhtmlpdf anymore?

chabad360:
Yes, I was thinking of getting it added to bench, but in that case I would think it might have to fully integrated into bench (i.e. multi tenancy would be seperate containers, supervisor wouldn't be used, etc.)

Is there a reason to put the MariaDB volume in ram (i.e. `/tmp`)? It doesn't need to be mapped to a folder to be persistent and putting it in `/tmp` actually deletes the db upon reboot.

Also, what is port `6787` for?

chabad360:
I will start working on replacing links with networks so it can be scaled, as well as integrating dockerize so that the `Procfile` and the `common_site_config` are easily configurable

chabad360:
update: now that I did some research and tests, it seems that using dockerize for the config files is more effort than it's worth. but there still may be other ways of doing the config setup (i.e. a simple entrypoint script that takes care of it).

achillesrasquinha:
As much as I like alpine for its small size, this is great for just a development setup. There are serious security concerns using alpine as your base image over debian

achillesrasquinha:
Could we consider flavoured frappe docker setups instead?

chabad360:
> I'm not thinking of bench based multi-tenency in production docker.

That's not what I meant, I meant that bench would manage docker.

But are you implying that we would run frappe raw (as in: without bench) in the container?

Thunderbottom:
closing due to inactivity.

PR #60, opened 2018-10-30
wkhtmltopdf binary qt patched for alpine linux

revant:
This is qt patched binary of wkhtmltopdf for alpine linux

This will be pulled to create docker image using `wget` in docker script


Thunderbottom:
closing since it is dependent on #59

PR #63, opened 2019-03-07
Fix and refractor a lot of things

chabad360:
This PR is **HUGE**, it adds a lot of different improvements and fixes, and overall shrinks the image and refractors in a lot of places.


chabad360:
I know that it seems like alot, but the biggest changes are to the README...

chabad360:
uhm, is there any activity going on here?

PR #72, opened 2019-08-13
chore: install wkhtmltox correctly

MorezMartin:
This way wkhtmltox is working with patched qt, last version avaible for debian stretch, the 0.12.5-1 doesn't seem to be supported yet

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,6 @@ services:
     volumes:
       - ./conf/mariadb-conf.d:/etc/mysql/conf.d
       - /var/lib/mysql
-        #      - ./db/var/lib/mysql for data persistence
     
     ports:
       - "3307:3306" # MariaDB Port
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,7 @@ services:
     volumes:
       - ./conf/mariadb-conf.d:/etc/mysql/conf.d
       - /var/lib/mysql
+        #      - ./db/var/lib/mysql for data persistence
     
     ports:
       - "3307:3306" # MariaDB Port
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
     volumes:
       - ./conf/mariadb-conf.d:/etc/mysql/conf.d
       - /var/lib/mysql
+        #      - ./db/var/lib/mysql for data persistence
```
chabad360: That line isn't needed, as it's already covered by 

```yaml
    - /var/lib/MySQL
```

MorezMartin: Better this way ?

chabad360: The less lines the better, and the less confusing

chabad360:
Perfect! Thank you.

I will have you know that you might be waiting a while till this gets merged. The maintainer seems to be on vacation...

MorezMartin:
Hello, no problem it is one of my first commits and i am happy to help, i
am not hurry, it is already working for me :)
Have a nice day,

MorezMartin:
Maybe it could be cool to add some classic fonts, i have been a little bit annoyed when i wanted to get cantarell working, i may update my Dockerfile
There is a repo there https://github.com/frappe/fonts
And for me apt-install fonts-cantarell but i have had problems
Maybe i will try to make a stack.yml file too
Have a nice day

chabad360:
Idk if you need them, it's a web app, so the site - upon loading - should automatically pull them

MorezMartin:
Yes, for previewing in browser but not for printing PDF

chabad360:
Oh right... Which fonts do you need?

Thunderbottom:
0.12.5 works fine now so could you please update this PR so we can merge it in?

MorezMartin:
Done

PR #76, opened 2019-11-02
stack.yml for swarm environment

MorezMartin:
Hey
Me again
This is a simple stack.yml file, it is not perfect and need to be improved
Sincerely yours
Martin

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,10 +19,10 @@ RUN apt-get update && apt-get install -y --no-install-suggests --no-install-reco
   && npm install -g yarn
 
 # Install wkhtmltox correctly
-RUN wget https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.4/wkhtmltox-0.12.4_linux-generic-amd64.tar.xz
-RUN tar xvf wkhtmltox-0.12.4_linux-generic-amd64.tar.xz
-RUN mv wkhtmltox/bin/wkhtmlto* /usr/bin/
-RUN ln -nfs /usr/bin/wkhtmltopdf /usr/local/bin/wkhtmltopdf
+RUN wget https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.4/wkhtmltox-0.12.4_linux-generic-amd64.tar.xz \
+  &&  tar xvf wkhtmltox-0.12.4_linux-generic-amd64.tar.xz \
+  &&  mv wkhtmltox/bin/wkhtmlto* /usr/bin/ \
+  && ln -nfs /usr/bin/wkhtmltopdf /usr/local/bin/wkhtmltopdf
 
 # Add frappe user and setup sudo
 RUN useradd -ms /bin/bash -G sudo frappe \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -16,15 +16,19 @@ ENV LC_ALL=en_US.UTF-8
...
+  software-properties-common sudo tk8.6-dev vim xfonts-75dpi xfonts-base wget wkhtmltopdf fonts-cantarell \
   && apt-get clean && rm -rf /var/lib/apt/lists/* \
   && curl https://deb.nodesource.com/node_10.x/pool/main/n/nodejs/nodejs_10.10.0-1nodesource1_amd64.deb > node.deb \
   && dpkg -i node.deb \
   && rm node.deb \
   && npm install -g yarn
 
+# Install wkhtmltox correctly
+RUN wget https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.5/wkhtmltox_0.12.5-1.stretch_amd64.deb
+RUN dpkg -i wkhtmltox_0.12.5-1.stretch_amd64.deb && rm wkhtmltox_0.12.5-1.stretch_amd64.deb
```
chabad360: Make this one run command

```Dockerfile
RUN wget https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.5/wkhtmltox_0.12.5-1.stretch_amd64.deb \
  && dpkg -i wkhtmltox_0.12.5-1.stretch_amd64.deb \
  && rm wkhtmltox_0.12.5-1.stretch_amd64.deb

```

This is because OCI container layers are immutable, so if it's not one `RUN` command you're not saving space...

stack.yml
```diff
@@ -0,0 +1,86 @@
...
+    driver_opts:
+      o: bind
+      type: none
+      device: /root/frappe/frappe_docker/frappe_docker/conf/redis-conf.d
+  frappe-bench:
+    driver: local
+    driver_opts:
+      o: bind
+      type: none
+      device: /root/frappe/frappe_docker/frappe_docker/frappe-bench
```
chabad360: Using the `local` driver will only allow the frappe service to "scale" on a server that has the path `/root/frappe/frappe_docker/frappe_docker/*`, hence it scales successfully because on every system other than the main one, the volume mounting fails. But, it will work to no end on a server that does have these folders, but if they're out of sync you're toast anyway.

This is all true if you're using regular folders, but if those folders are network shares of any kind (Gluster, NFS, SMB, etc.) then you'd want to specify those as the volume driver (if possible).

stack.yml
```diff
@@ -0,0 +1,86 @@
+version: '3'
```
chabad360: Move to version `3.7`.

chabad360:
For reasons other than my requested changes, I don't believe this is the right time for this PR to be merged. 

There are certain prerequisites for being production-ready, one of them being that you can be up and running with virtually no setup required (BTW trying to do a scripted frappe setup in a stack using `dbench` is very, very hard), but the current docker image is a pain to use and is definitely not production-ready.

Hence, as much as this is a great step in the right direction, this is just not yet worth merging if it's not including the necessary changes to make running `frappe_docker` on stack practical (but this is [on the way](https://github.com/chabad360/frappe_docker/compare/master...chabad360:stack)).

revant:
can you make the stack.yml for new production images?

also add a Recipe in Wiki for Swarm+Traefik+ERPNext
(Portainer or Swarmpit will be awesome)

We'll have devcontainers again later.

send things to develop https://github.com/frappe/frappe_docker/wiki/Contribution-Guidelines

PR #86, opened 2020-02-19
Workaround for https://github.com/docker/for-win/issues/1588

ceefour:


revant:
We'll have dev containers later again.

PR #84, opened 2020-02-10
feat: docker production images

revant:
Images based on #83 

Part of this PR only `edge` tag is built from develop branch. If accepted I'd suggest versions like this - https://hub.docker.com/r/castlecraft/frappe-socketio/tags. i.e v12, v12.2.1, v12.2.0; v11, v11.1.64, v11.1.63. Any other suggestions I'll change travis yaml accordingly

Travis CI Builds, 1 existing test job + 3 new build jobs

- https://travis-ci.org/revant/frappe_docker/builds/648331777

Docker Hub images auto pushed.

- https://hub.docker.com/r/revantnandgaonkar/frappe-socketio
- https://hub.docker.com/r/revantnandgaonkar/erpnext-python
- https://hub.docker.com/r/revantnandgaonkar/erpnext-assets

Frappe Team help needed,

- [x] In depth review for security, best practices, size, cleaner scripts.
- [x] add 2 environment variables `DOCKER_USERNAME` and `DOCKER_PASSWORD` for https://travis-ci.org/frappe/frappe_docker. Use official docker account and its token 
- [x] naming of images `frappe/*`
  - [x] create repository for `frappe/frappe-socketio`
  - [x] create repository for `frappe/erpnext-assets`
  - [x] create repository for `frappe/erpnext-worker`
  - [x] create repository for `frappe/frappe-worker`
  - [x] create repository for `frappe/frappe-assets`
- [x] Add Travis CI Cron Job to build images daily or weekly.

PR #87, opened 2020-02-19
Update frappe image

ceefour:
1. Update base image to `debian:buster-slim`, which updates Python to 3.7 as well
2. Update Node.js to 12.16.1

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Frappe Bench Dockerfile
 
-FROM debian:9.6-slim
+FROM debian:buster-slim
 LABEL author=frappé
 
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends locales \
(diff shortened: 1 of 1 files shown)


Thunderbottom:
Hey, we plan on completely replacing the current docker image with #84, feel free to suggest any changes there instead if you feel so.

ceefour:
@Thunderbottom can you reconsider merging this PR?   🙏

As @revant confirmed in https://github.com/frappe/frappe_docker/pull/84#issuecomment-590218034 , that PR is only for production use. So users like me still need the docker containers for development use, which the current one sufficiently providers, although needs some updating and maintenance as I proposed in this PR.

Thank you in advance. 🙏

Thunderbottom:
I'd suggest staying away from docker for development use. It doesn't make any sense to use docker for development to be honest.

ceefour:
@Thunderbottom Thank you for your suggestion. Despite obvious drawbacks, there are some people who want to use Docker for development. As @revant also pointed out, Microsoft also has [specific support for this use case in VS Code](https://code.visualstudio.com/docs/remote/containers). For example, Docker makes one good (but not _the only_) alternative for developing in a hybrid team with different OSes, since we can rely on the same base image.

If the concern is the burden of maintaining a variant of docker images for development purpose, I'm offering to maintain it in this or another repository under @frappe (perhaps @frappe_docker_dev?), which I definitely need for my own purpose, and also I hope will be helpful to others who have the same need. When @revant's production PR has been merged and replaced this repo entirely, I will try to mimic the structure as close as possible while maintaining developer-friendly features e.g. `bench`. I'd be happy if you can take my offer.

Thunderbottom:
You may do so if you feel like, but we do not plan on offering a docker for development. It just creates additional maintenance overhead which we'd like to avoid at this point. If needs be, you can setup VS Code to work with the production docker that is being worked on, since the only thing you'll be missing out inside it is bench.

PR #93, opened 2020-03-03
Add a build script for travis.yml

chabad360:
This PR adds a script that shrinks `travis.yml` by putting the common actions into a script (appropriately called `travis.sh`), which should help simplify writing an addition to `travis.yml`.

Issue #94, opened 2020-03-03
Improve Documentation


revant:
<strike>- Move Documentation to frappe website. Unnecessary to trigger builds on documentation changes</strike>
- [x] Describe each image separately in documentation
- [X] Section for single host docker-compose
- [ ] Section for multi node Docker swarm 
- [ ] Add and document Helm Charts for kubernetes based installation
- [x] Section to build production images based on frappe or erpnext images

ddavidebor:
Consider kubernetes instead of docker swarm, nobody is supporting docker swarm anymore

ddavidebor:
Also, please don't separate code documentation from the code, once you do it's impossible to enforce documentation improvements on pull request.

revant:
> Please consider Kubernetes instead of docker swarm, nobody is supporting docker swarm anymore

Kubernetes is expensive!

DO the easiest and cheapest I found. Right now it will need 3 nodes, 1 load balancer, 1 object storage and block storages for installed stateful apps. Can anyone fund me a cluster for 15 days?

I think most of the users are looking for "free" / "affordable" single-host installation with least complications and costs. 

From docker-compose it is easy to move to docker swarm. May be someone can take a step to make a single machine docker swarm cluster.

For Self hosting users with no intention to build SaaS Docker swarm may be sufficient.

Of course! everything in my plan is going on for Kubernetes!

Here is an existing Helm-Chart and "Job" Resources that help in creating/backing up and migrating sites. https://gitlab.com/castlecraft/erpnext_kubernetes/-/tree/master/kubernetes

ddavidebor:
For a simple and affordable single-host installation docker-compose is more than sufficient. if you don't have multiple nodes, there is no reason to use docker-swarm.

You can test kubernetes cluster on any single machine using minikube, you don't need a production grade cluster.

revant:
I make helm chart using Minikube.

I need to test it in production

ddavidebor:
If you can estimate the expenses for DO I'll see what I can do

revant:
> If you can estimate the expenses for DO I'll see what I can do

for reference
https://github.com/frappe/frappe_docker/issues/106

I've shared this with DO support and ERPNext Foundation.

ddavidebor:
#103 addresses point 2 " Describe each image separately in documentation" @revant

revant:
Closing this issue
For helm chart #106 
For docker swarm #118

PR #100, opened 2020-03-03
Refactor renaming erpnext-assets and frappe-assets to *-nginx 

### LINKED ISSUE #98: frappe-assets should be renamed
frappe-assets/erpnext-assets name is incorrect, as the container does not only serve static assets, but it also proxies request. This is very confusing. I recommend renaming "assets" to "webserver"

ddavidebor:
Fixes #98


revant:
If we're renaming
Let's wait for any other suggestions.

ddavidebor:
I liked your idead of -nginx, it's very clear. Everybody knows nginx is famous for two things: static assets serving and proxy

revant:
@Thunderbottom we need to make changes to docker hub repo.

Thunderbottom:
I think calling it `frappe-nginx` doesn't make sense, cause it's not just for the nginx service since it includes assets as well. I agree that calling it `frappe-assets` and `erpnext-assets` doesn't make sense either. I think `frappe-assets-server` and `erpnext-assets-server` sounds better? Let me know what you think, we can make changes to the dockerhub repo based on it.

rmehta:
`frappe-static` seems like a better option

Thunderbottom:
> `frappe-static` seems like a better option

yeah but since the assets docker also serves as a reverse proxy for serving the assets, i think specifying that it _serves_ the assets as well would make for a better name.

id say lets just take up a name and stick to it, and i'll list down what each container does in the documentation. i don't think naming anything should be such a big fuss ever.

rmehta:
Is the `assets` docker both revere proxy or only for asset serving? Do we have another docker for nginx ?

revant:
> Is the assets docker both revere proxy or only for asset serving? Do we have another docker for nginx ?

Yes it is both reverse proxy and serving assets

frappe-nginx

rmehta:
ok, then `frappe-nginx` is a better name

rmehta:
Let's also make this clear in the README. What each image does and how they are connected.

revant:
Can you close this and send it to develop branch?

I added this https://github.com/frappe/frappe_docker/wiki/Contribution-Guidelines

ddavidebor:
> Can you close this and send it to develop branch?
> 
> I added this https://github.com/frappe/frappe_docker/wiki/Contribution-Guidelines

I've changed the base.



> Let's also make this clear in the README. What each image does and how they are connected.

I have added a description to #94

revant:
@chabad360 travis needs fixed https://github.com/frappe/frappe_docker/pull/93

https://travis-ci.org/frappe/frappe_docker/jobs/658478555

chabad360:
This needs to be merged for the develop branch to build successfully. Can we stop bike-shedding please?

PR #104, opened 2020-03-04
Create a Wrapper script for installing custom apps

chabad360:
This PR adds two wrapper scripts for installing custom apps, which should significantly simplify the image creation process.

I have also implemented them into the erpnext image `Dockerfile`s to demonstrate their usage.


revant:
can you close this and send it to develop branch?

I just made https://github.com/frappe/frappe_docker/wiki/Contribution-Guidelines

chabad360:
> build/common/worker/install_app.sh should be moved ti build/common/commands where all such commands reside and possibly written in python

Actually neither, as it's not doing things post image build, it's only for use in the `Dockerfile`.

chabad360:
> install_app=['erpnext'] such things also need dynamic change.

Good catch, I'll have a look at that part of the equation next...

chabad360:
> `install_app=['erpnext']` such things also need dynamic change.

I disagree, I would say that because it's a modular container environment, we should assume that all apps in the `apps.txt` file should be installed. Unless we assume that this system is multi-tenant compatible (which it appears not to be), then we should make an environment variable called `INSTALL_APPS` that has a value of which apps you wish to install, comma delineated, and with a default value of `frappe`.

ddavidebor:
> Unless we assume that this system is multi-tenant compatible (which it appears not to be), in which case we should make an environment variable called `INSTALL_APPS` that has a value of which apps you wish to install, comma delineated, and with a default value of `frappe`.

I like this idea much better. The system is definitely multi-tenant compatible as it supports multiple domains and multiple sites.

revant:
It is multi-tenent compatible.

Allows multiple sites per deployment
Backup and migration by default runs on all sites

chabad360:
All's good now, merge conflicts resolved, and multi-tenancy is supported just set the `INSTALL_APPS` environment variable to whichever apps you want installed (i.e. `INSTALL_APPS="erpnext,custom_app"`,comma- delineated). 

All that's left is to update the docs...

chabad360:
Can someone do a code review and a test (as I'm not quite in a position to test this at the time being...), I'd like to make sure I don't miss anything and since I'm currently only available at night, I'm likely to miss something, or just completely botch something. So please, someone make sure I didn't miss anything.

.travis.yml
```diff
@@ -18,95 +18,95 @@ before_install:
...
         - docker build -t frappe/bench:latest -f build/bench/Dockerfile .
         - docker push frappe/bench:latest
-    - name: "Build Frappe python environment (edge)"
+    - stage: "Frappe (edge)"
       if: branch = develop AND type != pull_request
       script:
         - ./travis.py frappe --nginx --tag edge
         - ./travis.py frappe --nginx --tag develop--tag-only
-    - name: "Build Frappe nginx + static assets (edge)"
+    - stage: "Frappe (edge)"
```
revant: I'm confused?

Is this repeated?

/travis.py frappe --nginx --tag edge

revant:
./travis.py frappe --worker --tag edge 
 
somehow tagged the frappe/frappe-worker as latest?

.travis.yml
chabad360: We need to use stages so that the `frappe-*` images are built first, otherwise, we'll get the `install_app not found` error.

But each stage will run separately making builds take much longer, so if we name the stages the same thing, travis-ci will run them together.

revant: I think the frappe-nginx is built twice?

line 28 to line 37 .travis.yml

```
    - stage: "Frappe (edge)"
      if: branch = develop AND type != pull_request
      script:
        - ./travis.py frappe --nginx --tag edge
        - ./travis.py frappe --nginx --tag develop--tag-only
    - stage: "Frappe (edge)"
      if: branch = develop AND type != pull_request
      script:
        - ./travis.py frappe --nginx --tag edge
        - ./travis.py frappe --nginx --tag develop--tag-only
```

.travis.yml
```diff
@@ -18,95 +18,95 @@ before_install:
...
+    - stage: "Build Frappe bench development environment (latest)"
       if: branch = develop AND type != pull_request
       script:
         - docker build -t frappe/bench:latest -f build/bench/Dockerfile .
         - docker push frappe/bench:latest
-    - name: "Build Frappe python environment (edge)"
+    - stage: "Frappe (edge)"
       if: branch = develop AND type != pull_request
       script:
         - ./travis.py frappe --nginx --tag edge
```
revant: builds frappe-nginx

.travis.yml
```diff
@@ -18,95 +18,95 @@ before_install:
...
         - ./travis.py frappe --nginx --tag develop--tag-only
-    - name: "Build Frappe nginx + static assets (edge)"
+    - stage: "Frappe (edge)"
       if: branch = develop AND type != pull_request
       script:
         - ./travis.py frappe --nginx --tag edge
```
revant: also builds frappe-nginx

revant:
```
Step 2/2 : RUN install_app erpnext https://github.com/frappe/erpnext
 ---> Running in 2ee49e0e4415
error: switch `b' requires a value
ERROR: /home/frappe/frappe-bench/apps/erpnext is not a valid editable requirement. It should either be a path to a local project or a VCS URL (beginning with svn+, git+, hg+, or bzr+).                                                                                                                                                                                  
The command '/bin/sh -c install_app erpnext https://github.com/frappe/erpnext' returned a non-zero code: 1
```

revant:
https://travis-ci.org/github/frappe/frappe_docker/jobs/663136570

PR #184, opened 2020-04-09
Changed X-Frappe-Site-Name header to use value from `$host` instead o…

girip11:
…f `$http_host` in nginx configuration

ISSUE
-----
`$http_host` is used for setting header 'X-Frappe-Site-Name'
which adds port number to the header along with the host value. Frappe source app.py expects the header
value to contain only the host name and not the port number. So `$host` should be used instead of
`$http_host` to set the 'X-Frappe-Site-Name' header

`$http_host` vs `$host` in nginx
--------------------------------
`$http_host` contains the host name along with port number whereas `$host` contains only the host name in lowercase without the port number.

> `$host` - This variable is equal to line Host in the header of request or
> name of the server processing the request if the Host header is not available.
> This variable may have a different value from $http_host in such cases:
> * when the Host input header is absent or has an empty value,
> `$host` equals to the value of server_name directive;
> * when the value of Host contains port number, `$host` doesn't include
> that port number. $host's value is always lowercase since 0.8.17.
> - [$host vs $http_host stackoverflow](https://stackoverflow.com/questions/15414810/whats-the-difference-of-host-and-http-host-in-nginx)

From the frappe source file [app.py](https://github.com/frappe/frappe/blob/develop/frappe/app.py#L107), X-Frappe-Site-Name is used if its set.

```Python
site = _site or request.headers.get('X-Frappe-Site-Name') or get_site_name(request.host)
```

Since `$host` variable will never contain port number which is not the case with `$http_host`,
`$host` should be used for setting the header 'X-Frappe-Site-Name'. Otherwise we have issues with site serving.

Tested the above changes in compose as well as in swarm environment.
In compose, tested the site with host mapping of 80 and 8000. Works with both the host port mapping.

Tested with erpnext version - v12.5.2

Changes to be committed:
	modified:   build/common/nginx-default.conf.template

### DIFF
--- a/build/common/nginx-default.conf.template
+++ b/build/common/nginx-default.conf.template
@@ -29,9 +29,9 @@ server {
         proxy_http_version 1.1;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "upgrade";
-        proxy_set_header X-Frappe-Site-Name $http_host;
+        proxy_set_header X-Frappe-Site-Name $host;
         proxy_set_header Origin $scheme://$http_host;
-        proxy_set_header Host $host;
+        proxy_set_header Host $http_host;
 
         proxy_pass http://socketio-server;
     }
@@ -52,8 +52,8 @@ server {
     location @webserver {
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Proto $scheme;
-        proxy_set_header X-Frappe-Site-Name $http_host;
-        proxy_set_header Host $host;
+        proxy_set_header X-Frappe-Site-Name $host;
+        proxy_set_header Host $http_host;
         proxy_set_header X-Use-X-Accel-Redirect True;
         proxy_read_timeout 120;
         proxy_redirect off;


build/common/nginx-default.conf.template
```diff
@@ -29,9 +29,9 @@ server {
         proxy_http_version 1.1;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "upgrade";
-        proxy_set_header X-Frappe-Site-Name $http_host;
+        proxy_set_header X-Frappe-Site-Name $host;
         proxy_set_header Origin $scheme://$http_host;
-        proxy_set_header Host $host;
+        proxy_set_header Host $http_host;
```
revant, line 34: Is this change necessary?

https://github.com/frappe/bench/blob/develop/bench/config/templates/nginx.conf#L91

Here it is set to $host instead of $http_host?

build/common/nginx-default.conf.template
```diff
@@ -52,8 +52,8 @@ server {
     location @webserver {
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Proto $scheme;
-        proxy_set_header X-Frappe-Site-Name $http_host;
-        proxy_set_header Host $host;
+        proxy_set_header X-Frappe-Site-Name $host;
+        proxy_set_header Host $http_host;
```
revant, line 56: https://github.com/frappe/bench/blob/develop/bench/config/templates/nginx.conf#L91

girip11: ## Change 1
In the bench template I see `proxy_set_header X-Frappe-Site-Name {{ site_name }};`.

**X-Frappe-Site-Name** to `$host` is necessary because it always only contains the host name and never the port number, whereas $http_host contains whatever is set by the client in the HTTP Host header and  it can contain port number as well. If you have to follow bench template then `$host` is the right usage here for **X-Frappe-Site-Name**.

## Change 2
For the change in the line `proxy_set_header Host $http_host;`, it is not needed really. It can remain the same as the previous one which is `$host`. But I thought it will be better to set the **Host** header with the value whatever was passed from the client. `$http_host` contains the value of the **Host** header that was set by the client. You can check about the same of using `$http_host` vs `$host` to set Host header in this [issue](https://github.com/nginx-proxy/nginx-proxy/issues/763)


I hope the above explanation  answers your question.

build/common/nginx-default.conf.template
girip11: This [github issue](https://github.com/nginx-proxy/nginx-proxy/issues/763) on official nginx repo could answer your question as to why use $http_host to set the Host header.

build/common/nginx-default.conf.template
Thunderbottom: for change 2 i would say you should've let it be. it's better to not let user send the input as to what the `$http_host` would be is not known, but `$host` always gets set to `$server_name`, so you'd never face any issue regarding malformed requests sent to the server

Issue #183, opened 2020-04-03
Remote mysql hostname not supported 


asifkhanpathan:
Issue: Bug report

**Do the checklist before filing an issue:**
 - [ ] Can you replicate the issue?
 - [ ] Is this something you can debug and fix? Send a pull request! Bug fixes and documentation fixes are welcome

**Describe the bug** :chart_with_downwards_trend:
remote mysql hostname not supported when pass through environment variable  MARIADB_HOST like aws RDS database it gets its value from 'common_site_config.json' which is default to mariadb container.
`docker exec -it \
                                       -e "SITE_NAME=test2" \
                                       -e "DB_ROOT_USER=admin" \
                                       -e "MYSQL_ROOT_PASSWORD=$MYSQL_ROOT_PASSWORD" \
                                       -e "ADMIN_PASSWORD=$ADMIN_PASSWORD" \
                                       -e "INSTALL_APPS=erpnext" \
                                   -e "MARIADB_HOST=test.ajdkkkdh.us-east-2.rds.amazonaws.com" \
                                       localhost_erpnext-python_1 docker-entrypoint.sh new
`
output Traceback
```
`Traceback (most recent call last):
  File "/home/frappe/frappe-bench/env/lib/python3.7/site-packages/pymysql/connections.py", line 583, in connect
    **kwargs)
  File "/opt/bitnami/python/lib/python3.7/socket.py", line 707, in create_connection
    for res in getaddrinfo(host, port, 0, SOCK_STREAM):
  File "/opt/bitnami/python/lib/python3.7/socket.py", line 752, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -2] Name or service not known

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/frappe/frappe-bench/commands/new.py", line 59, in <module>
    main()
  File "/home/frappe/frappe-bench/commands/new.py", line 25, in main
    reinstall=False,
  File "/home/frappe/frappe-bench/apps/frappe/frappe/commands/site.py", line 75, in _new_site
    source_sql=source_sql, force=force, reinstall=reinstall, db_type=db_type, db_host=db_host, db_port=db_port, db_password=db_password, no_mariadb_socket=no_mariadb_socket)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/installer.py", line 35, in install_db
    setup_database(force, source_sql, verbose, no_mariadb_socket)
[...]

PR #195, opened 2020-04-17
Use Docker volumes for storing sites

tbolender:
When setting up ERPNext in our environment, I stumbled upon the usage of the `sites` folder as volume. This can cause problems regarding permissions depending on your operating system and/or configuration. I received permission errors preventing the creation of a new site.

Especially in production setup, this is not a nice solution. I propose the usage of volumes for the site storage as well. If really required, access to the folder is (at least under Linux) still possible.

P.S. Thanks a lot for your work, it made setting up of a page pretty straight forward!


revant:
>  I received permission errors preventing the creation of a new site.
If the sites directory is owned by uid 1000 and gid 1000, generally first user on linux then there won't be any permission problem

I don't mind it being Docker volume instead of bind mount. It is easy for users to edit sites config directly from host if it's bind mount.

Additional best practice has to be explained to "docker run" new container with shared volume to edit config
(AWS rds needs change to config json). This should also be part of documentation then?

Making it docker volume is first step to multi host volumes

tbolender:
@ddavidebor Updated. I also fixed all backup and restore instructions. I missed it before.

@revant I am not sure whether I correctly get what for instructions you mean. To launch a one-time container to edit the config in the console? Which image should we use for this?

revant:
> To launch a one-time container to edit the config in the console? Which image should we use for this?

check this issue: https://github.com/frappe/frappe_docker/issues/183

user has to add `"rds_db": 1` to `common_site_config.json`. For that you need to access the file.

in case of bind mount it is as easy as editing the file from host.

in case of docker volume, another container with editor has to run with sites-vol mounted to edit common_site_config.json

README.md
```diff
@@ -250,7 +250,7 @@ Environment Variables
     -e "SECRET_ACCESS_KEY=secret_access_from_provider" \
     -e "ENDPOINT_URL=https://region.storage-provider.com" \
     -e "BUCKET_DIR=frappe-bench-v12" \
-    -v ./installation/sites:/home/frappe/frappe-bench/sites \
+    -v sites-vol:/home/frappe/frappe-bench/sites \
```
revant, line 253:
```suggestion
    -v <project-name>_sites-vol:/home/frappe/frappe-bench/sites \
```

volumes generally have prefix when created as part of docker-compose

revant:
How do I setup AWS RDS in this setup? I need to edit `sites/common_site_config.json` for that

https://github.com/frappe/frappe/wiki/Using-Frappe-with-Amazon-RDS-(or-any-other-DBaaS)

Common question in Issues and on forum.

README.md
tbolender: Fixed in one other place as well.

PR #201, opened 2020-04-20
Merge branch 'develop'

Thunderbottom:
merging branch to trigger a build on erpnext version-12 for a fix that wasn't released in the tag

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -8,5 +8,12 @@
   "dockerComposeFile": "./docker-compose.yml",
   "service": "frappe",
   "workspaceFolder": "/workspace/development",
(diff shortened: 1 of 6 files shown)


revant:
This breaks semver.

ddavidebor:
I write really unprofessional commit messages sometimes...

revant:
I think this breaks semver:
https://github.com/frappe/erpnext/compare/v12.7.0...version-12

Current latest branch `version-12` has latest tag `v12.7.0` in case of ERPNext. (At the time of writing)
There is no patch version and `version-12` branch is ahead by 5 commits

I think this is clean right now:
https://github.com/frappe/frappe/compare/v12.5.0...version-12

PR #216, opened 2020-04-27
fix: README

revant:
single command to deploy locally
env file placed where docker-compose will be executed


blaggacao:
Thankf for your quick fix!

PR #227, opened 2020-05-03
Update branch version while installing app

maheshk2k:
Both bench init and get-app need to be on the same version branch. 
Mismatch results into errors while installing app


maheshk2k:
Request reviewers to review and accept the change to fix branch version mismatch.

PR #232, opened 2020-05-06
Fix socketio error in reverse proxy setup

### LINKED ISSUE #231: Reverse Proxy: Error in socketio requests - it refers to sitename instead of hostname
## Description of the issue
In my dockerized setup with a reverse proxy in front of ERPNext, I have error message in the logfile in the container `frappe-socketio`. It can't access the host with the name of the ERPNext site name.

On my system I install ERPNext with the official docker images. There is already a nginx included but I opted to connect the erpnext-nginx to a Traefik reverse proxy to fit to the overall architecture.

# Setup:
* I set up everything on version-11 images
* I created a site called `site1.local` in erpnext
* The Traefik reverse proxy listens to the public host name of the server (e.g. https://erp.example.com)
* Traefik modifies the `host` header of the request from erp.example.com to site1.local

--> Then I'm able to access my ERPNext instance through Traefik (this is the expected behavior)

1. Setup dockerized ERPNext
2. Setup traefik
3. Connect traefik with erpnext-nginx and set the header `host` to the value of the site

In my case I use Traefik 2.2 with this labels on the erpnext-nginx container - development, without ssl:

```
  erpnext-nginx:
    image: frappe/erpnext-nginx:${VERSION}
    restart: on-failure
    labels:
      - "traefik.enable=true"
      - "traefik.docker.network=proxy"
      - "traefik.http.routers.erpnext_insecure.rule=Host(`erpnext.example.com`)"
      - "traefik.http.routers.erpnext_insecure.entrypoints=http"
      - "traefik.http.routers.erpnext_insecure.service=erpnext_insecure@docker"
      - "traefik.http.services.erpnext_insecure.loadbalancer.server.port=80"
      - "traefik.http.middlewares.hostname.headers.customrequestheaders.host=site1.local"
      - "traefik.http.routers.erpnext_insecure.middlewares=hostname"
```

### Observed result
* In the logfile I get the error as shown in the stacktrace
* socketio do use the sitename instead of the hostname of the request
### Expected result
* let socketio use the hostname of the request

### Stacktrace / full error message if available

```
 Error: getaddrinfo ENOTFOUND site1.local
     at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:66:26) {
   errno: -3008,
   code: 'ENOTFOUND',
   syscall: 'getaddrinfo',
   hostname: 'site1.local',
   response: undefined
 }
```

paulator:
Trying to fix #231 with setting the Header `origin` to `$http_referer`.

`$http_referer` seems to hold the URL which the request comes from  (e.g. http://erp.example.com) 

Socketio extracts the value from the `origin` header (see here: https://github.com/frappe/frappe/blob/253109e2f9413c5e5851f50ebd54260b0623fbf7/socketio.js#L288)

In my case it fixes the problem but I'm not sure, if it affects something else...

### DIFF
--- a/build/common/nginx-default.conf.template
+++ b/build/common/nginx-default.conf.template
@@ -30,7 +30,7 @@ server {
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "upgrade";
         proxy_set_header X-Frappe-Site-Name $host;
-        proxy_set_header Origin $scheme://$http_host;
+        proxy_set_header Origin $http_referer;
         proxy_set_header Host $http_host;
 
         proxy_pass http://socketio-server;


revant:
I built the images locally, I found this error as soon as any user logs in.

```
frappe-socketio_1        | Unable to join chat room. Error: getaddrinfo ENOTFOUND test.localhost
```

everything seem to work in chat.

revant:
We do need to merge this on master for v11 images to reflect changes. can you confirm if things work with edge images?

paulator:
Thank you for your quick response and your review.

After some intensive testing I discovered, that the chat does not automatically show the new messages. It shows only messages after refreshing the screen. But I don't see any error message in the logs.

Even worse, I get a couple of error messages like this one, when I click through ERPNext documents:

```
 Error: Not Found
     at Request.callback (/home/frappe/frappe-bench/apps/frappe/node_modules/superagent/lib/node/index.js:871:15)
     at IncomingMessage.<anonymous> (/home/frappe/frappe-bench/apps/frappe/node_modules/superagent/lib/node/index.js:1114:20)
     at Stream.emit (events.js:315:20)
     at Unzip.<anonymous> (/home/frappe/frappe-bench/apps/frappe/node_modules/superagent/lib/node/unzip.js:53:12)
     at Unzip.emit (events.js:315:20)
     at endReadableNT (_stream_readable.js:1223:12)
     at processTicksAndRejections (internal/process/task_queues.js:84:21) {
   status: 404,
```

I'm afraid that this "fix" is not sufficient for that it was supposed to.

The only possibility I have so far is to give the site the same name as the Traefik (public url) is.

revant:
I'll merge anything that doesn't break.

I'll revert it if it causes issue later.

I generally try to keep nginx.conf similar to one in frappe/bench. Easy to maintain and copy changes from core to here.

paulator:
Ok, thx.

Do you have an idea how to fix that issue? I really want stick to that scenario. 

It would be easy for me to setup a dev/test/prod environment with the same site name and various URL names.

revant:
You need multiple urls to point at same site?

There's jinja2 template that generates nginx conf for native frappe setup. I've translated it to allow dynamic site names.

https://github.com/frappe/bench/blob/develop/bench/config/templates/nginx.conf

check how additional domains are added in site_config.json and those are rendered into config.

https://frappe.io/docs/user/en/bench/guides/adding-custom-domains

We need this part dynamic. Nginx container that is running cannot be restarted on adding site or domain name.

Also I feel the domains must be there in site_config.json I guess they tell frappe werkzeug which site to serve. I've to confirm. There must be some part in Frappe to validate which site to serve for incoming request's host

paulator:
Hi,

well, actually I don't need multiple urls for a site in my setup. I also tried setting up the host_name item in the site_config.json. But it didn't change anything.

The basic requirement is the following:

- I want to backup a production ERPNext installation (e.g. hostname: `erp.example.com`, sitename: `erp.example.com`)
- Then I want to restore the site 1:1 on the test server (e.g. hostname: `erp.test.example.com`, sitename: `erp.example.com`)

The setup with reverse proxy is:

production:
`reverse-proxy (listens erp.example.com) --forward--> erpnext (site: erp.example.com)`

test:
`reverse-proxy (listens erp.test.example.com) --forward(*)--> erpnext (site: erp.example.com)`

(*) in the forwarder of the proxy I rewrite the HOST header to:  `erp.example.com`.

The idea with multiple URLs could be a sufficient solution, then I don't need to rewrite the header and use the second URL instead. I'll give it a try and report back here.

revant:
There's an hack for backup / restore that may help.

the command `restore-backup` will restore any site mounted/located at `/home/frappe/backups`. Refer Directory structure in readme.

Backup and you'll get directory database and files for erp.example.com

Create/Rename the files and directory (file and directory structure in readme) to erp.test.example.com and restore it on test server.

PR #254, opened 2020-06-10
Multibench support in devel environment

africlouds:
The current development container only exposes ports 8000 and 9000. It is not possible to have multiple bench since each bench needs a port. A Dockerfile was added to development to expose 5 ports (8000 to 8005 and 9000 to 9005). 

docker-compose was also updated accordingly

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,2 @@
+FROM frappe/bench:latest
+EXPOSE 8000-8005 9000-9005 6787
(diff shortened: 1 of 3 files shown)


africlouds:
Nothing breaks in the VS Code workflow?

revant:
We'll fix if anything breaks!

I'll try today.

revant:
https://github.com/frappe/frappe_docker/pull/256 continue discussion here.

PR #256, opened 2020-06-11
feat(frappe/bench): allow 5 benches

revant:
- Instead of building one more image locally, updated the bench image.
- update to readme as per changes

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,2 +1,0 @@
-FROM frappe/bench:latest
-EXPOSE 8000-8005 9000-9005 6787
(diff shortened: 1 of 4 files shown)


revant:
@africlouds Why did you make new image to build locally? any specific reason?

africlouds:
> @africlouds Why did you make new image to build locally? any specific reason?

I was thinking 1) One bench would be enough in production 2) Giving the developer a full Dockerfile would give more power to install extra packages in a custom image

But thinking of it on 2), it is better to do that in an extra docker-compose.yml file located in workspace/development untracked by git

PR #286, opened 2020-06-25
fix: restore encryption key on site restore

Thunderbottom:
if encryption key exists for a site, it will now be restored to the site config while restoring site

### DIFF
--- a/build/common/commands/restore_backup.py
+++ b/build/common/commands/restore_backup.py
@@ -216,7 +216,7 @@ def main():
                 db_password=site_config.get('db_password'),
             )
             make_site_dirs()
-            restore_database(files_base, site)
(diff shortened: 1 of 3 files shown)


build/common/commands/restore_backup.py
```diff
@@ -180,8 +190,9 @@ def main():
         latest_backup = max(backups).strftime(DATE_FORMAT)
         files_base = os.path.join(backup_dir, site, latest_backup, '')
         files_base += latest_backup + '-' + site_slug
+        site_config_path = os.path.join(backup_dir, site, 'site_config.json')
```
revant, line 193: there are 2 types of file names.

for v12 onwards:

```
20200625_233122-mysite_localhost-site_config_backup.json
```

v11:

```
site_config.json
```

PR #295, opened 2020-07-01
feat: ping additional services using doctor

revant:
doctor --ping-service postgres:5432 -p mariadb:3306


ddavidebor:
@revant perhaps this should be documented?

revant:
> @revant perhaps this should be documented?

It needs fix, split of string by character (":") is not handled properly and may throw value error.

I'll add section in site operations for this.

This is meant to be used in kubernetes health probes.

PR #299, opened 2020-07-06
feat: profit off HTTP caching

sunhoww:
Without a `.build` file, the assets served get a random `build_version` on every request. Benefits from `Cache-Control` headers are not utilized. 
https://github.com/frappe/frappe/blob/cac7cc402fa8e28120284ae49a6cff0fb8681c38/frappe/www/desk.py#L85-L91

I added the command to `touch .build` in the _Dockerfile_. But I don't know whether it would be more appropriate to do this in _install_app.sh_.

### DIFF
--- a/build/erpnext-nginx/Dockerfile
+++ b/build/erpnext-nginx/Dockerfile
@@ -12,6 +12,7 @@ FROM frappe/frappe-nginx:${GIT_BRANCH}
 COPY --from=0 /home/frappe/frappe-bench/sites/ /var/www/html/
 COPY --from=0 /rsync /rsync
 RUN echo "erpnext" >> /var/www/html/apps.txt
+RUN touch -m /var/www/html/.build
 
 VOLUME [ "/assets" ]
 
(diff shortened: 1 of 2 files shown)


revant:
Make the command part of existing layer instead of adding one more layer. Less layers to download.

I'm trying to figure out best place to add `.build` file such that custom apps should have least rituals to follow.

revant:
https://github.com/frappe/frappe/blob/develop/rollup/build.js#L108-L111

revant:
It seem it's not that simple.

1. The `.build` file needs to be in `frappe-bench/sites`. The python code you shared looks for it in sites path.
2. `sites` is already a volume. Adding a .build file there will keep it there forever, without updating the timestamp.
3. entrypoint needs to touch this .build file if the image got updated.

sunhoww:
> 2. sites is already a volume. Adding a .build file there will keep it there forever, without updating the timestamp.

You're right. My bad. I failed to take this into account.

> 3. entrypoint needs to touch this .build file if the image got updated.

`touch` from here will mean the build version will update on container restarts. Which is fine - at least better than the random string alternative. I was thinking of putting the `touch` command in `/rsync` itself from _install_app.sh_. Which essentially ends up with the same behavior.

Should I close this PR and create a new one with the changes in _docker-entrypoint.sh_ instead?

revant:
/rsync as name suggests will do rsync.

Add new thing that makes it known to one who reads the code, that it touches .build.

refer how auto migrate works. It checks the version change in py container.

Here in nginx container there won't be py available.
I'll suggest simple bash script that does

Check timestamp of .build.
Check timestamp of assets in image
If assets are newer than .build, re create it.

Close this PR after you read this comment.

PR #307, opened 2020-07-13
feat: touch .build

sunhoww:
Continued from #299.

This sets the `.build` modified time to the newest modified times of all the folders in `/assets`. Every time any of the assets are rebuilt, their modified time updates which in turn does the same to `.build`.

Sorry couldn't update this sooner.

### DIFF
--- a/build/frappe-nginx/docker-entrypoint.sh
+++ b/build/frappe-nginx/docker-entrypoint.sh
@@ -13,6 +13,8 @@ rsync -a --delete /var/www/html/assets/frappe /assets
 
 chmod -R 755 /assets
 
+touch /var/www/html/sites/.build -r $(ls -td /assets/* | head -n 1)
+
 if [[ -z "$FRAPPE_PY" ]]; then
     export FRAPPE_PY=0.0.0.0
 fi


revant:
Does touch changes modified time everytime it runs?
If so every time the container starts it'll modify the file.

In scenarios where container/service is scaled, every container that starts will modify the file.

Ideas:

Store ERPNext and Frappe version in .build file
Check containing version before touch.
Save this file in python container where it can have full py environment to save

sunhoww:
I started out by having the touch command inside an `if` conditional that checks whether `.build` exists or whether `.build` is older than the newest asset folder. Later on, it seemed simpler for `touch` to just change the modified time to be equal to modified time of the newest folder in `/assets`.

> Does touch changes modified time everytime it runs?

I also thought this might happen. However, it didn't. It's a good thing you are using `rsync -a` to copy over the assets, because this command preserved the modified time of the source. So essentially, `.build` is modified only when rebuilt images are used, and not every time containers restart.

revant:
Try it on develop images after they are built.

sunhoww:
I'd been using the custom `docker-entrypoint.sh` on a production build with custom apps for a few days. It's been holding up. I'll verify with develop.

Issue #308, opened 2020-07-11
[bug] Error installing ERPNext v12.10.1 on AWS RDS MariaDB


ceefour:
## Description of the issue

I have enabled `"rds_db": 1` in `common_site_config.json`:

```
(env) root@frappe-bench-0001-erpnext-worker-d-847966697-fwkwc:/home/frappe/frappe-bench/sites# cat common_site_config.json 
{
    "db_host": "xxx.ap-southeast-1.rds.amazonaws.com",
    "db_port": 3306,
    "rds_db": 1,
    "redis_cache": "redis://frappe-bench-0001-erpnext-redis-cache:13000",
    "redis_queue": "redis://frappe-bench-0001-erpnext-redis-queue:12000",
    "redis_socketio": "redis://frappe-bench-0001-erpnext-redis-socketio:11000",
    "socketio_port": 9000
}
```

## Context information (for bug reports)

* Kubernetes 1.18
* frappe/erpnext-worker v12.10.1
* AWS RDS MariaDB 10.4.13

## Steps to reproduce the issue

1. Create AWS RDS MariaDB database
2. Install ERPNext v12.10.1 on Kubernetes using helm chart
3. Error

### Observed result

```
ERROR 1054 (42S22) at line 1: Unknown column 'ERROR (RDS): SUPER PRIVILEGE CANNOT BE GRANTED OR MAINTAINED' in 'field list'
ERROR 1396 (HY000) at line 1: Operation ALTER USER failed for '_9a6f28ddcbb1acb1'@'%'
ERROR 1044 (42000) at line 1: Access denied for user 'root'@'%' to database '_9a6f28ddcbb1acb1'
```

### Expected result

Successful.

### Stacktrace / full error message

```
$ kubectl create -n erpnext -f add-lovia-site-job.yaml 
...

$ kubectl logs -f -n erpnext create-erp-lovia-life-2knwd
Attempt 1 to connect to xxx.ap-southeast-1.rds.amazonaws.com:3306
Attempt 1 to connect to frappe-bench-0001-erpnext-redis-queue:12000
Attempt 1 to connect to frappe-bench-0001-erpnext-redis-cache:13000
Attempt 1 to connect to frappe-bench-0001-erpnext-redis-socketio:11000
Connections OK
Created user _9a6f28ddcbb1acb1
Created database _9a6f28ddcbb1acb1
Granted privileges to user _9a6f28ddcbb1acb1 and database _9a6f28ddcbb1acb1
[...]

PR #312, opened 2020-07-15
fix: mariadb grants for aws rds

revant:
For #308


build/common/commands/new.py
```diff
@@ -96,6 +96,11 @@ def main():
         command = mysql_command + [f"GRANT ALL PRIVILEGES ON `{db_name}`.* TO '{db_name}'@'%'; FLUSH PRIVILEGES;"]
```
Thunderbottom, line 96: I think this command shouldn't run if `rds_db` is set since RDS does not allow running `GRANT ALL`

Thunderbottom: so that'll fail, instead we could have something like:

```py
grant_privileges = RDS_PRIVILEGES if config.get(RDS_DB) else "ALL PRIVILEGES"
command = mysql_command + [f"GRANT {grant_privileges} ON `{db_name}`.* TO '{db_name}'@'%'; FLUSH PRIVILEGES;"]
```

and inside `constants.py`

```py
RDS_PRIVILEGES = "SELECT, INSERT, UPDATE, DELETE, CREATE, DROP, INDEX, ALTER, CREATE TEMPORARY TABLES, CREATE VIEW, EVENT, TRIGGER, SHOW VIEW, CREATE ROUTINE, ALTER ROUTINE, EXECUTE, LOCK TABLES"

PR #323, opened 2020-08-02
fix(erpnext-nginx): remove unused docker-entrypoint script

### LINKED ISSUE #321: extra file creating confusion
why there is docker-entrypoint.sh in build/erpnext-nginx folder when it is already in build/frappe-nginx and being used everywhere.

revant:
fixes #321


premkanish:
thanks!

Issue #325, opened 2020-08-04
Restore backup fails when using erpnext with AWS RDS 


freewayz:
## Description of the issue
When restoring an existing backup site, it fails if your database is hosted with AWS RDS, because of this line https://github.com/frappe/frappe_docker/blob/develop/build/common/commands/restore_backup.py#L230 due to AWS RDS not supporting `GRANT ALL` as part of the default `SHOW GRANTS ;` priviledges

## Steps to reproduce the issue

1.  Setup an AWS RDS mariadb
2. Backup your local/existing erpnext site
3. Restore backup using this command https://github.com/frappe/frappe_docker/blob/develop/docs/site-operations.md#restore-backups

### Observed result
```Overwrite site demo.example.come
Extract Database GZip for site demo.example.com
Something went wrong:
return code: 1
stdout:
b''

stderr:
b"ERROR 1044 (42000) at line 1: Access denied for user 'user'@'%' to database '_649xxxxxx'\n"
```

### Expected result
  - Should restore backuped site
### Stacktrace / full error message if available

```
(paste here)
```

PR #326, opened 2020-08-04
fix: grant privileges for rds when restoring backup

freewayz:

Fixes https://github.com/frappe/frappe_docker/issues/325


freewayz:
@revant can you help give this PR a look, we in the middle of a migration for a client.

PR #327, opened 2020-08-04
fix: grant privileges for rds when restoring backup

revant:
merge to master to trigger production builds


revant:
`docker build -t frappe/frappe-worker:develop -f build/frappe-worker/Dockerfile .` failing

PR #332, opened 2020-08-08
build: Allow cached layer for install_packages git python2

ceefour:
Related to #320

Since `install_packages git python2` is run by `install_app.sh`, better make it part of the previous Docker layer, saves time for slow internet users (like me).

### DIFF
--- a/build/erpnext-nginx/Dockerfile
+++ b/build/erpnext-nginx/Dockerfile
@@ -2,6 +2,9 @@ ARG NODE_IMAGE_TAG=12-prod
 ARG GIT_BRANCH=develop
 FROM bitnami/node:${NODE_IMAGE_TAG}
 
+# make it faster (cached layer) for next "/install_app" step
+RUN install_packages git python2
+
 ARG GIT_BRANCH
 COPY build/erpnext-nginx/install_app.sh /install_app
 


revant:
Then remove it from install_app.sh script?

It adds more ritual steps for new users who were trying to build images.

ceefour:
I don't think it needs to be removed from the script.

In this PR it is flexible, the Dockerfile will preinstall these packages. 

However in case user modifies the Dockerfile, the script will install these packages anyway. (if did not exist) 

So in both cases end result is same, but different build time needed.

revant:
If users is modifying this dockerfile on fork, let this not be PR?

ceefour:
User is only modifying the repository url for the app to be installed.

But the previous lines including this PR usually will be left untouched.

Thunderbottom:
I think we can get rid of the base image that is being used to build `erpnext-nginx`. The second image that we use, `frappe-nginx`, already has git installed. So instead of doing the current workflow like:

```
bitnami/node => build assets => copy to frappe-nginx
```

we can directly do:

```
frappe-nginx as base => build and copy assets
```

that'll require some changes to the `install_app` script, but other than that shouldn't need anything.

revant:
@ceefour 
What does not feel right, the command runs in 2 different layers.

Anyway it doesn't matter, final image is built from minimum layers without unnecessary packages.

@Thunderbottom frappe-nginx is also a dual layered image to optimize size. Second/final layer doesn't have git.

revant:
> So in both cases end result is same, but different build time needed.

This ERPNext file is intended to used as part of automated CI process.

If at all anyone wan't to experiment locally with cache or other base images they can change it as per need.

I don't think additional steps to explain should be part of CI build.

For now I'll close this.

Issue #229, opened 2020-05-05
Can not update Website_Theme, Missing Node...


AlvaroMolano:
## Description of the issue
Node.js problems to perform style generation.
Edit, update or create a website theme style is not working.
FileNotFound: [Error2] No such file or directory 'node':'node'
## Context information (for bug reports)
frappe_docker installed version12 
New site install

## Steps to reproduce the issue
1. Use a clean frappe_docker setup
2. Create new site
3. Go to Website module, and edit standard theme. Save changes. Or create new theme and try to save changes.

![image](https://user-images.githubusercontent.com/3089876/81084997-e26b4000-8ef6-11ea-9c5a-31588b95f8c8.png)

### Observed result RIGHT IMAGE
Message error: Node not found.
Some tools running on node might not be working.
### Expected result LEFT IMAGE
Changes are saved as in a bench install without docker.

### Stacktrace / full error message if available
Log from <project>_erpnext-python_1 container
```
Traceback (most recent call last):
  File "/home/frappe/frappe-bench/apps/frappe/frappe/desk/form/save.py", line 22, in savedocs
    doc.save()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 273, in save
    return self._save(*args, **kwargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 296, in _save
    self.insert()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 230, in insert
    self.run_before_save_methods()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 893, in run_before_save_methods
    self.run_method("validate")
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 794, in run_method
    out = Document.hook(fn)(self, *args, **kwargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 1064, in composer
    return composed(self, method, *args, **kwargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 1047, in runner
    add_to_return_value(self, fn(self, *args, **kwargs))
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 788, in <lambda>
    fn = lambda self, *args, **kwargs: getattr(self, method)(*args, **kwargs)
[...]

PR #341, opened 2020-09-02
fix(erpnext-nginx): add node-sass as production dependency

revant:
https://github.com/frappe/frappe_docker/issues/229#issuecomment-685849850

@MarekPikula can you test this if it works?


MarekPikula:
Sure, I'll look into it tomorrow.

MarekPikula:
Is it now available in any particular tag, or should I build it myself?

revant:
It is available in all tags in a while, merged to master.

Issue #346, opened 2020-08-24
Liveness probe performance issue


MarekPikula:
## Description of the issue

I deployed a ERPNext Helm chart on in-house K8S cluster. Recently, after upgrade to newer chart (from 1.0.0 to 1.0.14), I've been noticing significant, constant CPU usage. It turned out that it is caused by liveness probe execution for scheduler and all workers, ie. `doctor.py` script. It's not that critical when the server is idle, but during heavy load it makes a significant performance impact.

First point is that exactly the same script is executed for multiple containers and doesn't really check if a given container is live, but if all backend services are live. Another is that it's a Python script, which by itself isn't particularly optimal. Ie. every 5s 4 instances of `docker-endpoint.sh` are executed, which execute `su` (what spams in sys log as well), initiate Python env, load Python interpreter with all the required libraries and execute a script, which, in my understanding after briefly reading the code, basically does a TCP liveness check.

## Context information (for bug reports)

Kubernetes cluster (v1.18.8) deployed on dedicated server on Fedora 33 with erpnext-nginx in version v12.10.1, erpnext-worker v12.10.1, socketio v12.8.4, helm chart v1.0.14.

## Steps to reproduce the issue

1. Deploy ERPNext Helm chart.
2. Watch CPU usage as `doctor.py` is being executed every 5s for scheduler and all 3 workers.

### Observed result

High CPU usage every 5s on a task that should be as optimised as possible (ie. liveness probe).

### Expected result

No significant CPU load during liveness check.

### Stacktrace / full error message if available

Not relevant.

MarekPikula:
Yeah, I started working on this, but unfortunately for the last week I had more urgent assignments in my company and since I'm the only SysOp there I cannot continue work on this for a week or so. I have it high in my backlog though, since it's really impacting server's performance.

revant:
doctor.py is complex. it's checking postgres host in helm chart.

https://github.com/frappe/helm/blob/65c9b96519255bead14a9f4abfb59e58c1a7c546/erpnext/templates/deployment-scheduler.yaml#L47

Just a thought: For helm chart/deployments use env variable based probes instead of doctor.py

revant:
changes for helm chart https://github.com/frappe/helm/pull/57

MarekPikula:
That's very clean solution indeed. Thanks for the work.

PR #348, opened 2020-09-15
fix: health check

revant:
#346


revant:
adding wait-for-it.sh is 8.0K +

should it be installed as a package instead? https://tracker.debian.org/pkg/wait-for-it

Thunderbottom (approved):

PR #352, opened 2020-09-22
Use new version variables in compose files

tbolender:
When updating our ERPNext deployment, I noticed that you made progress on Kubernetes support. It seems that you introduced more specific version variables for this, namely `FRAPPE_VERSION` and `ERPNEXT_VERSION`.

However, I noticed that the docker-compose files still use the old `VERSION` variable. I replaced them with the correct one so that they work with the new version variables.

### DIFF
--- a/env-example
+++ b/env-example
@@ -1,4 +1,5 @@
-VERSION=edge
+ERPNEXT_VERSION=edge
+FRAPPE_VERSION=edge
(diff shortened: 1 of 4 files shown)


revant:
Tests depend on those files,. Tests will also need some changes to variable

tbolender:
This should be fixed now

PR #357, opened 2020-10-01
feat: add drop site command to worker

sahil28297:
Command to drop site from bench
1. deletes the site folder and stores its backup in the archived_sites folder in the sites directory (change the path of `ARCHIVE_SITES_PATH` in constants.py if you want to store it elsewhere)
2. deletes the site database from mariadb.


revant:
@sahil28297 can you add this to site-operations.md

PR #368, opened 2020-10-14
feat(frappe-worker): update user query for mariadb galera

revant:
mariadb galera helm chart doesn't allow root@% to ALTER mysql.user


Thunderbottom (approved):
LGTM!

PR #381, opened 2020-11-24
fix: make new command compatible with MySQL 8

everyx:
Starting with MySQL 8, we no longer can (implicitly) create a user using the `GRANT` command. we should Use `CREATE USER` instead followed by the GRANT statement.


revant:
Thanks for the PR!

PR #383, opened 2020-11-24
fix: install app failed when APP_NAME is not same as the one containe…

everyx:
install app failed when `APP_NAME` is not same as the one contained in `APP_REPO`


Thunderbottom (approved):
Could replicate, looks good to me!

PR #401, opened 2021-01-04
Non root password expiry

revant:
- fixes password expiry error for old images
- update NodeJS to latest minor and patch lts release


revant:
https://travis-ci.com/github/frappe/frappe_docker

Travis not starting any job

PR #419, opened 2021-02-17
Allow for configuring the realip module for nginx to pass the proper IP/Scheme to frappe

Xaroth:

Currently, frappe-nginx (and subsequently erpnext-nginx) don't report the proper IP/scheme when placed behind a traefik instance.

These changes configure the [nginx real-ip module](http://nginx.org/en/docs/http/ngx_http_realip_module.html) to allow the X-Forwarded-For and X-Forwarded-Proto headers sent by traefik (and other loadbalancers) to be used. 

This enables 3 new environment variables that can be passed:
- UPSTREAM_REAL_IP_ADDRESS  to configure the trusted IP range (to avoid IP spoofing). By default this is 127.0.0.1, but for docker this can be set to 172.17. 0.0/16 (non-swarm) or 10.0.0.0/8 (swarm), or whatever loadbalancer/proxy IP range is being used.
- UPSTREAM_REAL_IP_RECURSIVE to configure the real_ip_recursive directive. This is mostly useful when behind nexted loadbalancers (i.e. AWS ELB => Traefik => frappe-nginx). This is off by default.
- UPSTREAM_REAL_IP_HEADER to configure which header will be used to obtain the original IP address from. By default we use X-Forwarded-For, but in some cases one might want to use X-Real-IP.

This change also adds the X-Forwarded-For and X-Forwarded-Proto headers for the socket-io endpoint, so that it is also aware of the real origin IP/protocol.

### DIFF
--- a/docs/docker-swarm.md
+++ b/docs/docker-swarm.md
@@ -142,6 +142,7 @@ services:
   erpnext-nginx:
     image: frappe/erpnext-nginx:${ERPNEXT_VERSION?Variable ERPNEXT_VERSION not set}
     environment:
+      - UPSTREAM_REAL_IP_ADDRESS=10.0.0.0/8
       - FRAPPE_PY=erpnext-python
       - FRAPPE_PY_PORT=8000
       - FRAPPE_SOCKETIO=frappe-socketio
--- a/build/common/nginx-default.conf.template
+++ b/build/common/nginx-default.conf.template
@@ -6,6 +6,12 @@ upstream socketio-server {
     server ${FRAPPE_SOCKETIO}:${SOCKETIO_PORT} fail_timeout=0;
 }
 
+# Parse the X-Forwarded-Proto header - if set - defaulting to $scheme.
+map $http_x_forwarded_proto $proxy_x_forwarded_proto {
+    default $scheme;
+    https https;
+}
+
 server {
     listen 80;
     server_name $http_host;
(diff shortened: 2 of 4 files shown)


revant:
How does it translate to Kubernetes?

Any suggestions for changes in https://github.com/frappe/helm?

revant (changes requested):
Add this to appropriate documentation.
There is also a separate file to document environment variables that are not mandatory or set by default.

Xaroth:
> How does it translate to Kubernetes?
> 
> Any suggestions for changes in https://github.com/frappe/helm?

The changes itself should apply to any setup where frappe-nginx is being served from behind a loadbalancer. I left it pretty generic to cover as much use-cases as possible.

The only configuration required is setting the appropriate env vars (if/when needed), so that would have to be picked up by the helm charts.

docs/docker-swarm.md
```diff
@@ -142,6 +142,7 @@ services:
   erpnext-nginx:
     image: frappe/erpnext-nginx:${ERPNEXT_VERSION?Variable ERPNEXT_VERSION not set}
     environment:
+      - UPSTREAM_REAL_IP_ADDRESS=10.0.0.0/8
```
revant, line 145: UPSTREAM_REAL_IP_RECURSIVE is off by default? does it need to be on here?

revant (changes requested):
If I merge this as it is, It won't affect the current working images right? defaults are "off".

I'll merge this after your improvements/comments on this review.

I'll need your help in updating the frappe/helm chart. I can add this to the deployment with defaults set. We need to update the documentation for setting this when loadbalancer is https://github.com/kubernetes/ingress-nginx (which is part of current documentation)

Alternatively, AWS ALB is also used.

docs/docker-swarm.md
Xaroth: This is only needed when you have more than one proxy server before nginx. Since there is only traefik, it can be left as off.

Xaroth:
The defaults are set so normal behavior will not be changed. Nginx will not process real ips because it only trusts itself, and not process the X-Forwarded-For header.

Xaroth:
wrt the Helm chart, most values for UPSTREAM_REAL_IP_ADDRESS will vary on the setups people use. The easiest is to assign it to the private address range that k8s uses, but this can differ between setups. The same goes for _RECURSIVE and _HEADER, though X-Forwarded-For is by far the most common you see when it comes to loadbalancers/proxies, so the default will suffice in most cases.

My take would be to expose the 3 new variables as Helm values, so that users can set them themselves to match their setup.

PR #422, opened 2021-02-20
Rootless containers by default

cortopy:

This PR introduces rootless containers by default

## Problem

Running docker containers as root is [extremely unsafe](https://americanexpress.io/do-not-run-dockerized-applications-as-root/) and should be avoided at all times unless no other options are available.

When deploying erpnext to a kubernetes cluster I was forced to run all containers as root with all privileges allowed. I noticed that there has been some thought about this and there is an environment variable called `$RUN_AS_ROOT` that allows to perform some actions as a non-root user. However, the use of this variable is sub-optimal because:

- It's not well documented. Indeed, the docker-compose examples that any new developer may use don't even mention the variable
- It's not really effective. Some commands have it and others don't.
- It creates a large amount of conditional branches that are hard to maintain

Crucially, I found that setting `$RUN_AS_ROOT` as 0 would not work because the `checkConnection` function does not include it and so one can end up with a very messy environment where some actions are performed as root and some aren't. 

In my case my problem was double:
- A very inconsistent use of root
- Privilege escalation all over
- Problems with permissions for pretty much everything (is the owner root or the frappe user?)

## Proposed solution

This PR does the following:

1. Removes all privilege escalation statements with `su`. Privilege escalation should only be used for system actions and executing bench commands does seem to require such a high level of permissions
2. Remove all `chown` actions. Again, this is considered a dangerous action during runtime. All file permissions should be handled at build time or by the container runtime (Docker, kubernetes, etc.) and never at runtime. It's much easier to use tried-and-tested approaches. For example, in kubernetes one can use `fsGroup`, an init container, etc.

## Also included

One of the `chown` actions is against the logs directory. This made me realised that there is one extra volume that wasn't declared before with the result that logs are lost when the container is recreated. There is now a new logs volume

### DIFF
--- a/build/common/worker/docker-entrypoint.sh
+++ b/build/common/worker/docker-entrypoint.sh
@@ -46,8 +46,8 @@ function configureEnv() {
 }
 
 function checkConnection() {
-  su frappe -c ". /home/frappe/frappe-bench/env/bin/activate \
-    && python /home/frappe/frappe-bench/commands/check_connection.py"
+  . /home/frappe/frappe-bench/env/bin/activate \
+    && python /home/frappe/frappe-bench/commands/check_connection.py
 }
 
 function checkConfigExists() {
@@ -68,9 +68,6 @@ if [[ ! -e /home/frappe/frappe-bench/sites/apps.txt ]]; then
   find /home/frappe/frappe-bench/apps -mindepth 1 -maxdepth 1 -type d -printf '%f\n' | sort -r > /home/frappe/frappe-bench/sites/apps.txt
 fi
 
-# Allow user process to create files in logs directory
-chown -R frappe:frappe /home/frappe/frappe-bench/logs
-
 # symlink node_modules
 ln -sfn /home/frappe/frappe-bench/sites/assets/frappe/node_modules \
   /home/frappe/frappe-bench/apps/frappe/node_modules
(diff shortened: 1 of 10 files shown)


build/common/worker/docker-entrypoint.sh
```diff
@@ -46,8 +46,8 @@ function configureEnv() {
 }
 
 function checkConnection() {
-  su frappe -c ". /home/frappe/frappe-bench/env/bin/activate \
```
cortopy, line 49: this privilege escalation with `su` forces all containers to be run with a very dangerous permission to escalate to root

build/common/worker/docker-entrypoint.sh
```diff
@@ -68,9 +68,6 @@ if [[ ! -e /home/frappe/frappe-bench/sites/apps.txt ]]; then
   find /home/frappe/frappe-bench/apps -mindepth 1 -maxdepth 1 -type d -printf '%f\n' | sort -r > /home/frappe/frappe-bench/sites/apps.txt
 fi
 
-# Allow user process to create files in logs directory
-chown -R frappe:frappe /home/frappe/frappe-bench/logs
```
cortopy, line 72: this path is not declared as a volume and can produce data loss

revant:
I'll have to test locally.

migration failed in tests, and edge site is showing "updating". https://travis-ci.com/github/frappe/frappe_docker/builds/217710707#L5149

I reported frappe error https://github.com/frappe/frappe/issues/12445.

revant:
Tested locally.
Works as expected.

Issue #431, opened 2021-02-27
ImportError: cannot import name '_new_site' from 'frappe.commands.site'


Goeppert:
Hi there,  Thanks for putting together this resource.  

I followed your instructions for setting up a single bench install (https://github.com/frappe/frappe_docker/blob/develop/docs/single-bench.md)

And I got the following error:

 ```
ubuntu@eg-7-ssd-bhs1:~/frappe_docker$ sudo docker logs kevin419_site-creator_1 -f
  Attempt 1 to connect to mariadb:3306
  Attempt 2 to connect to mariadb:3306
  Attempt 3 to connect to mariadb:3306
  Attempt 1 to connect to redis-queue:6379
  Attempt 1 to connect to redis-cache:6379
  Attempt 1 to connect to redis-socketio:6379
  Connections OK
  Traceback (most recent call last):
    File "/home/frappe/frappe-bench/commands/new.py", line 5, in <module>
      from frappe.commands.site import _new_site
  ImportError: cannot import name '_new_site' from 'frappe.commands.site' (/home/frappe/frappe- bench/apps/frappe/frappe/commands/site.py)
```

Mohamedmahrous995:
the same issue with me , did you fix it ?

smartekIT:
I confirm, the same with me for the last 2 days and no way to create the site.

Thunderbottom:
Seems like an import path change in the upstream, have sent a pull request to fix this.

Thunderbottom:
Seems like the import path for the `_new_site()` function has changed in the upstream, have sent a pull request to fix this.

Mohamedmahrous995:
i have added the from frappe.installer.site import _new_site but still not working when i connect to browser it gives me internal server error !!!!

Thunderbottom:
Not sure why that would happen. Seems like a separate issue unrelated to this one. If it persists after the pull request is merged, please feel free to open a new issue describing the internal server error

smartekIT:
mate even after the merge, the creator container still gives the same error:'
ImportError: cannot import name '_new_site' from 'frappe.commands.site' (/home/frappe/frappe- bench/apps/frappe/frappe/commands/site.py)

Thunderbottom:
Figured it out, my bad. I hope I can use lack of sleep as an excuse for this one.

PR #433, opened 2021-02-28
fix: change import path for new_site command

Thunderbottom:
`_new_site` has been moved to `installer.py` in frappe/frappe#12447

closes: #431

### DIFF
--- a/build/common/commands/new.py
+++ b/build/common/commands/new.py
@@ -2,7 +2,7 @@ import os
 import frappe
 import semantic_version
 
-from frappe.installer import _new_site, update_site_config
+from frappe.installer import update_site_config
 from constants import COMMON_SITE_CONFIG_FILE, RDS_DB, RDS_PRIVILEGES
 from utils import (
     run_command,
@@ -10,6 +10,16 @@ from utils import (
     get_site_config,
     get_password,
 )
+
+# try to import _new_site from frappe, which could possibly
+# exist in either commands.py or installer.py, and so we need
+# to maintain compatibility across all frappe versions.
+try:
+    # <= version-{11,12}
+    from frappe.commands.site import _new_site
+except ImportError:
+    # >= version-13 and develop
+    from frappe.installer import _new_site
 
 
 def main():
--- a/build/common/commands/new.py
+++ b/build/common/commands/new.py
@@ -2,8 +2,7 @@ import os
 import frappe
 import semantic_version
 
-from frappe.commands.site import _new_site
-from frappe.installer import update_site_config
+from frappe.installer import _new_site, update_site_config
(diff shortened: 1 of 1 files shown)


Mohamedmahrous995:
Traceback (most recent call last):
  File "/home/frappe/frappe-bench/commands/new.py", line 5, in <module>
    from frappe.commands.site import _new_site
ImportError: cannot import name '_new_site' from 'frappe.commands.site' (/home/frappe/frappe-bench/apps/frappe/frappe/commands/site.py)
it gives me this error after adding the line code from frappe.installer.site import _new_site
to new.py

Thunderbottom:
>it gives me this error after adding the line code from frappe.installer.site import _new_site to new.py

Did you replace the old import or did you just add this line to the file? Cause the prior still will cause an import error. Also, the CI seems to be working fine now so I'm pretty sure that the issue's fixed.

smartekIT:
> > it gives me this error after adding the line code from frappe.installer.site import _new_site to new.py
> 
> Did you replace the old import or did you just add this line to the file? Cause the prior still will cause an import error. Also, the CI seems to be working fine now so I'm pretty sure that the issue's fixed.

replacing the old import still gives the same error. nothing changed

3xitsharp:
It should be `from frappe.installer import _new_site` without ".site" ?

https://github.com/frappe/frappe/blob/develop/frappe/commands/site.py#L33

Thunderbottom:
Correct, my bad. Weird how the CI still works though.

revant (commented):
Does this change backward compatible upto v11?

If not then we need to check `frappe.__version__`?

Thunderbottom:
Doesn't look like it is backwards compatible. Let me fix it up a bit.

revant:
> Correct, my bad. Weird how the CI still works though.

CI only fails on exit 1.

curl, w3m not configured to fail. Many times frappe patches fail and we still need to merge docker fixes

Thunderbottom:
We can use `curl --fail` in those places maybe. From the `man` page:

```
-f, --fail
              (HTTP)  Fail  silently (no output at all) on server errors. This
              is mostly done to better enable scripts etc to better deal  with
              failed  attempts.  In  normal cases when an HTTP server fails to
              deliver a document, it  returns  an  HTML  document  stating  so
              (which  often  also describes why and more). This flag will pre‐
              vent curl from outputting that and return error 22.

              This method is not fail-safe and there are occasions where  non-
              successful response codes will slip through, especially when au‐
              thentication is involved (response codes 401 and 407).
```

revant:
we need to skip fails. many times migrations fail. I think we have 1 failing patch even right now. So let it be skipped. I check the test and just scan through that it should only silently fail on migrations.

for other tests we can `--fail` or similar? If you feel something can be improved do send the changes.

Issue #437, opened 2021-03-19
add gevent in env


revant:
https://github.com/frappe/frappe_docker/blob/5121e2710653c8df513e953a21e8d1ee8d3c2bde/build/common/worker/docker-entrypoint.sh#L97

docker starts gunicorn with gthread worker-class. (seems part of gunicorn)

https://docs.gunicorn.org/en/stable/settings.html?highlight=gevent#worker-class

gevent can be installed and used as option.

Found here : https://discuss.erpnext.com/t/fastapi-vs-werkzeug/72785/5

revant:
refer PR for further discussion

PR #438, opened 2021-03-20
feat(frappe-worker): option to set gevent as gunicorn worker-class

revant:
#437


revant:
Anyone else can also comment and review

revant:
Merging this, I may have to make changes to helm chart as well, I'll test it on my prod first.

PR #441, opened 2021-03-26
feat: Add Site name and Host as env variables for NGINX config template

mr-karan:
The custom NGINX image doesn't allow to set static Host/Site-Name headers. This is required if the app is exposed to an AWS ALB which can't send these headers for doing the healthchecks. Since the healthchecks fail, the target groups are not activated, thus giving 502.

For such environments, the host header has to be hardcoded inside the NGINX config. This change is backwards compatible since it fallbacks to using the original variables inside the config template, only overrides if the user gives them as environment variables in their `docker-compose` configs.

### DIFF
--- a/build/frappe-nginx/docker-entrypoint.sh
+++ b/build/frappe-nginx/docker-entrypoint.sh
@@ -45,6 +45,15 @@ if [[ -z "$UPSTREAM_REAL_IP_HEADER" ]]; then
     export UPSTREAM_REAL_IP_HEADER="X-Forwarded-For"
 fi
 
+if [[ -z "$FRAPPE_SITE_NAME_HEADER" ]]; then
+    export FRAPPE_SITE_NAME_HEADER="\$host"
+fi
+
+if [[ -z "$HTTP_HOST" ]]; then
+    export HTTP_HOST="\$http_host"
+fi
+
+
 envsubst '${FRAPPE_PY}
     ${FRAPPE_PY_PORT}
     ${FRAPPE_SOCKETIO}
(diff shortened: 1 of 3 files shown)


build/frappe-nginx/docker-entrypoint.sh
```diff
@@ -45,13 +45,24 @@ if [[ -z "$UPSTREAM_REAL_IP_HEADER" ]]; then
     export UPSTREAM_REAL_IP_HEADER="X-Forwarded-For"
 fi
 
+if [[ -z "$FRAPPE_SITE_NAME_HEADER" ]]; then
+    export FRAPPE_SITE_NAME_HEADER="\$host;"
```
revant, line 49: `;` semi-colon should be removed from both variable value, it already exists in template file.

it causes template to result into extra semi-colon in `default.conf` file

```
        proxy_set_header X-Frappe-Site-Name $host;;
        proxy_set_header Host $http_host;;
```

Note the extra semi-colon.

revant (changes requested):
- Add the details and defaults for the env variables in docs, https://github.com/frappe/frappe_docker/blob/develop/docs/environment-variables.md
- Not related to this PR, there can be a variable that altogether bypass the template to default.conf generation. If set, the nginx conf can be mounted as configmap and the container can use that directly.

mr-karan:
> Not related to this PR, there can be a variable that altogether bypass the template to default.conf generation. If set, the nginx conf can be mounted as configmap and the container can use that directly.

Yep, that would be awesome :+1: I'll work on a separate PR for this.

revant:
Tests passed.

In the test I found that it is not able to ping to new created site, It shows 503,

https://travis-ci.com/github/frappe/frappe_docker/builds/221302952#L5267

I'll test manually and then merge.

revant (approved):

revant:
503 was because first migration failed in test, that didn't revert common_site_config.maintenance_mode

PR #462, opened 2021-04-16
docs: consolidate installation instructions

vasanthv16:
Updated doc - mainly moved Frappe/ERPNext installation instructions from frappe/bench to frappe_docker. PR contains two commits to make diff-ing easier. First commit is a blind move of content from bench to frappe_docker. Second commit has some minor updates to the README.

Please review this PR in conjunction with https://github.com/frappe/bench/pull/1153.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,19 @@
 |---------|-----------------------------------------------------------------------------------------------------------------------------|
 | Master  | [![Build Status](https://travis-ci.com/frappe/frappe_docker.svg?branch=master)](https://travis-ci.com/frappe/frappe_docker) |
 
-## Getting Started
+## Table of Contents
 
-### Try in Play With Docker
+- [Try in Play With Docker](#try-in-play-with-docker)
+- [Installation](#installation)
+   - [Docker Installation](#docker-installation)
+      - [Setting up Pre-requisites](#setting-up-pre-requisites)
+      - [Development Setup](#development-setup)
+      - [Production Setup](#production-setup)
+   - [Easy Install Script](#easy-install-script)
+   - [Manual Installation](#manual-installation)
+- [Contribution](#contributing)
+
+## Try in Play With Docker
 
 <a href="https://labs.play-with-docker.com/?stack=https://raw.githubusercontent.com/frappe/frappe_docker/develop/tests/pwd.yml">
   <img src="https://raw.githubusercontent.com/play-with-docker/stacks/master/assets/images/button.png" alt="Try in PWD"/>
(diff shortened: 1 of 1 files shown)


revant:
The ERPNext installation need not be moved here from
frappe/bench!

If you need any changes in this docs send it here. Adding easy install and manually install here is confusing.

@ddavidebor very much cares about docs and its integrity. Waiting for comments.

revant:
Why Docker documentation is here?

It is here because it is maintained along with the source code. It's expected to update docs when code is updated.

If it's moved to frappe_docs it may become out of sync.

Also frappe_docs is today's location, I've seen locations change and documentation is lost during the change. I lost oauth2 docs. Raffael cared to rewrite the docs related to Oauth2 and that's why they exist today.

In ideal world where documentation doesn't get lost it could be in one place.

Related issue https://github.com/frappe/frappe_docs/issues/79

Any ideas to solve this fundamentally?

GitHub action or Travis job to map docs/*.md from frappe_docker to frappe_docs on every change to docs directory here. Just thinking out loud.

vasanthv16:
Thanks for the review.

There are really different ways to maintain documentation - in one consolidation master location, or each repo having documentation relevant to it. I have my opinion based on my limited knowledge, but I will leave it to you folks to decide whats best.

I saw documentation in two different places (bench and frappe_docker), and I could see things be a bit out of whack between the two. So I thought there is benefit in consolidating it.

On hindsight, I should have probably moved just the docker installation content to frappe/frappe_docker and leave easy install script and manual installation in frappe/bench. I could do that, if there is agreement.

revant:
> There are really different ways to maintain documentation - in one consolidation master location, or each repo having documentation relevant to it. I have my opinion based on my limited knowledge, but I will leave it to you folks to decide whats best.

I don't decide how docs will be generated. The basic thing I can do is keep markdown docs in frappe_docker/docs

> I saw documentation in two different places (bench and frappe_docker), and I could see things be a bit out of whack between the two. So I thought there is benefit in consolidating it.
> 

Yes there is!

I don't know why frappe/bench is the place where ERPNext installation docs are?
frappe_docker has no bench at all. It only has bench_helper.py which is part of frappe framework.

I don't think anyone from official team uses docker or even plan to use it. Branding, marketing, re-branding, domain location change, there are many variables that only core team decides upon and during migrations frappe_docker and frappe/helm will be last thing considered as it is not used internally. I (bots) have to keep track on their activities to keep containerization updated.

This repo is managed by community contributions and I'd not like losing anything from here.

> On hindsight, I should have probably moved just the docker installation content to frappe/frappe_docker and leave easy install script and manual installation in frappe/bench. I could do that, if there is agreement.

Docker installation content is already in frappe_docker/docs as single-bench, multi-bench or cluster installs?

https://github.com/frappe/frappe_docker/blob/develop/README.md#production-setup

No one will land on frappe/frappe_docker in the first place! I wanted to check flow for users exploring ERPNext. I took following steps. 

1. went to erpnext.com or erpnext.org
2. directed to https://github.com/frappe/erpnext#full-install, here docker is not mentioned and people generally go to bench or download VM image. Docker should be mentioned here as a link to repo, just like the link to bench repo
3. directed to https://github.com/frappe/bench (priority is on easy install, docker installation was added later and becomes outdated fast as the real docs keep updating here on frappe/frappe_docker)

After a while, I just started keeping the docs in docs directory under our control.

vasanthv16:
Thanks for the details on how things work. That definitely gives me more context on a lot of things.

When I was looking to install ERPNext, I followed a similar path as you mentioned, starting with erpnext.com. I tried VM image, but it was for v12. I then tried easy install on the latest version of Ubuntu and ran into some issue (don't recall the issue exactly). I finally stumbled upon docker and that has worked well for me.

Personally, I think docker is a better way to install since everything is self contained. There is no confusion about which version of Ubuntu to use, etc.

As for next steps, here is what I am thinking.

(1) Update my frappe/bench and frappe/frappe_docker PRs to pull out only docker installation steps from frappe/bench into frappe/frappe_docker, and leave easy install and manual install in frappe/bench.
(2) Create a new PR to update https://github.com/frappe/erpnext#full-install to refer to Docker installation, and point to the README in frappe/frappe_docker. I think we can/should recommend docker ahead of VM and easy install options.

What do you think?

revant:
> pull out only docker installation steps from frappe/bench into frappe/frappe_docker

May be keep a link to this repo. Someone may land on frappe/bench and get directed here.

I don't think this repo needs any change (no need to copy steps from frappe/bench to here). It was all copied from here to frappe/bench and this repo got improved after that.

Again if you feel the README.md of frappe_docker needs to be improved make those changes. Just copying will copy old text.

> Create a new PR to update https://github.com/frappe/erpnext#full-install to refer to Docker installation, and point to the README in frappe/frappe_docker. I think we can/should recommend docker ahead of VM and easy install options.

+1

revant:
https://github.com/frappe/erpnext/pull/25455

Issue #123, opened 2020-03-07
update dev container


ddavidebor:
You can assign this issue to me
update development container to use debian as close as possible to the one of the production containers

revant:
- Lock the python 3 versions to the ones in official frappe and erpnext travis tests
- Keep 2 versions of nodejs (v10 & v12) with nvm. ERPNext v11 and v12 development support is already there. If you are able to add ERPNext v10 support it'll be awesome.
- Install nvm in user's $HOME and with user ownership. It allows users to do `npm i <package> -g` without `sudo`.

If anyone can try windows 10 + docker desktop and vs code containers it will help in improving the image

ddavidebor:
Opened #133 to complete before this issue to have more insight
Opened #132 to simplify work a little bit on stuff like this one for production containers

ddavidebor:
@revant I'm not great with virtualenvs if you have suggestions on how to implement your point I'm listening

> * Lock the python 3 versions to the ones in official frappe and erpnext travis tests

ddavidebor:
Opened issue https://github.com/frappe/frappe_docker/issues/137 regarding VSCode testing in Windows, as it's better to do it after the various PR are merged and the container updated

ddavidebor:
> * Lock the python 3 versions to the ones in official frappe and erpnext travis tests
I'm not great with virtualenvs, do you have recommendations?
> * Keep 2 versions of nodejs (v10 & v12) with nvm. ERPNext v11 and v12 development support is already there. If you are able to add ERPNext v10 support it'll be awesome.
Done ! #135 
> * Install nvm in user's $HOME and with user ownership. It allows users to do `npm i <package> -g` without `sudo`.
Done ! #135

PR #476, opened 2021-05-21
feat: improve build

revant:
use nodejs v14
allow build args to pass repo and branch
add WORKER_CLASS under erpnext-python in docker-compose.yml

### DIFF
--- a/build/bench/Dockerfile
+++ b/build/bench/Dockerfile
@@ -1,5 +1,9 @@
 # Frappe Bench Dockerfile
 FROM debian:buster-slim
+
+ARG GIT_REPO=https://github.com/frappe/bench.git
+ARG GIT_BRANCH=develop
+
 LABEL author=frappé
 
 RUN apt-get update -y && apt-get install  \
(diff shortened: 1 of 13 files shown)


Thunderbottom (approved):
LGTM!

build/bench/Dockerfile
```diff
@@ -87,13 +91,12 @@ RUN bash -c "bench --version"
 # !!! UPDATE NODEJS PERIODICALLY WITH LATEST VERSIONS !!!
 # https://nodejs.org/en/about/releases/
 # https://nodejs.org/download/release/latest-v10.x/
-# https://nodejs.org/download/release/latest-v12.x/
-# https://nodejs.org/download/release/latest-v13.x/
-ENV NODE_VERSION=12.20.0
-ENV NODE_VERSION_FRAPPEV11=10.23.0
+# https://nodejs.org/download/release/latest-v14.x/
+ENV NODE_VERSION=14.17.0
+ENV NODE_VERSION_FRAPPEV11=10.24.1
```
PeterDaveHello, line 96: node.js v10 is EOL already, ref: https://github.com/nodejs/Release/#end-of-life-releases, maybe should also update it to v12 (consider to keep two versions of node.js here, ref: https://github.com/frappe/frappe_docker/issues/123#issuecomment-596211931)? I can send a pull request if it make sense to you.

revant: I think it's needed for old v11 development.
ERPNext V11 won't work without nodejs v10.

PeterDaveHello: Oh, okay, I got it. Thanks.

PR #486, opened 2021-06-08
feat: allow custom frappe branch during asset image build

revant:
closes https://github.com/revant/erpnext_feature_board/issues/15

### DIFF
--- a/build/erpnext-nginx/Dockerfile
+++ b/build/erpnext-nginx/Dockerfile
@@ -6,6 +6,7 @@ FROM node:${NODE_IMAGE_TAG}
 
 ARG GIT_REPO=https://github.com/frappe/erpnext
 ARG GIT_BRANCH=develop
+ARG FRAPPE_BRANCH=${GIT_BRANCH}
 
 COPY build/erpnext-nginx/install_app.sh /install_app
 RUN chmod +x /install_app && \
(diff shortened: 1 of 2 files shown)


Alchez (approved):

build/erpnext-nginx/Dockerfile
```diff
@@ -14,7 +15,7 @@ RUN chmod +x /install_app && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 
-RUN /install_app erpnext ${GIT_REPO} ${GIT_BRANCH}
+RUN /install_app erpnext ${GIT_REPO} ${GIT_BRANCH} ${FRAPPE_BRANCH}
```
Thunderbottom, line 18: shouldn't this be

```suggestion
RUN /install_app erpnext ${GIT_REPO} ${FRAPPE_BRANCH}
```

revant: No. Frappe branch is additional arg now.
For regular builds behaviour doesn't change.
For PRs to have base frappe too build assets the branch need to be correct.
Branch same as fork's branch causes frappe to not get cloned during build.

Ideally the install_app script needs improvement.

Probably we should use python for the script, we'll get argparse and build better cli for install_app

PR #490, opened 2021-06-21
refactor: remove exports, add bash flags healthchecks

pratikbalar:
Add bash flags and removed 
- `errexit` - better for fast fail of healthchecks
- `allexport` - export vars by default
- `nolog` - not to add in history (optional)

### DIFF
--- a/build/common/worker/healthcheck.sh
+++ b/build/common/worker/healthcheck.sh
@@ -1,31 +1,31 @@
 #!/bin/bash
 set -ea
 
-function getRedisUrl() {
-  cat ${1} | grep $2 | awk -v word=$2 '$word { gsub(/[",]/,"",$2); print $2}' | tr -d '\n' | sed 's|redis://||g'
+function getUrl() {
+  cat ${1} | grep $2 | awk -v word=$2 '$word { gsub(/[",]/,"",$2); print $2}' | tr -d '\n'
 }
 
 COMMON_SITE_CONFIG_JSON='/home/frappe/frappe-bench/sites/common_site_config.json'
 
 # Set DB Host and port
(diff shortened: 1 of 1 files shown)


revant (approved):

build/common/worker/healthcheck.sh
```diff
@@ -1,28 +1,29 @@
 #!/bin/bash
+set -o errexit -o allexport -o nolog
```
Thunderbottom, line 2: isn't `nolog` ignored by default?

can we instead make this:

```suggestion
set -ea
```

pratikbalar: Umm, yeah
> isn't nolog ignored by default?

sure, I just go with the verbose 
` set -ea`

build/common/worker/healthcheck.sh
```diff
@@ -1,28 +1,29 @@
...
 
-export COMMON_SITE_CONFIG_JSON='/home/frappe/frappe-bench/sites/common_site_config.json'
+COMMON_SITE_CONFIG_JSON='/home/frappe/frappe-bench/sites/common_site_config.json'
 
 # Set DB Host and port
-export DB_HOST=`cat $COMMON_SITE_CONFIG_JSON | awk '/db_host/ { gsub(/[",]/,"",$2); print $2}' | tr -d '\n'`
-export DB_PORT=`cat $COMMON_SITE_CONFIG_JSON | awk '/db_port/ { gsub(/[",]/,"",$2); print $2}' | tr -d '\n'`
+DB_HOST=`cat $COMMON_SITE_CONFIG_JSON | awk '/db_host/ { gsub(/[",]/,"",$2); print $2}' | tr -d '\n'`
```
Thunderbottom, line 7: since these `awk` commands are repeatedly being used, let's turn it into a function in the file for simplicity.

pratikbalar: done

PR #496, opened 2021-06-28
ci: build arm64 images

revant:
Build arm64 images

Grouped all the frappe builds in a stage and erpnext builds in stage following frappe build stage.


revant:
We'll get back to this later when builds are approx 30 min

Issue #448, opened 2021-04-01
pymysql.err.OperationalError: (1045, "Access denied for user 'root'@'172.19.0.6' (using password: YES)")


Natetronn:
## Description of the issue

My issue is similar to this one, although on production (that issue is over a year old, didn't want to necro open it, and it was noted to open new issue for production; also, not using bench, as this is production so, not all is relevant; although I did try some of the solutions there, without success ): https://github.com/frappe/frappe_docker/issues/239

## Steps to reproduce the issue

1. After what appeared to be a successful setup of Docker et al, following the [Docker installation instructions](https://github.com/frappe/bench#docker-installation)
2. Get to the point of starting things up by running the following command:
```
docker exec -it \                          
    -e "SITE_NAME=$SITE_NAME" \
    -e "DB_ROOT_USER=$DB_ROOT_USER" \
    -e "MYSQL_ROOT_PASSWORD=$MYSQL_ROOT_PASSWORD" \
    -e "ADMIN_PASSWORD=$ADMIN_PASSWORD" \
    -e "INSTALL_APPS=erpnext" \
    mysitecom_erpnext-python_1 docker-entrypoint.sh new
```

### Observed result
Access denied for user roor error:
`pymysql.err.OperationalError: (1045, "Access denied for user 'root'@'172.19.0.6' (using password: YES)")`

### Expected result
Expected root to have access to mariadb

### Stacktrace / full error message if available

```
docker exec -it \                          
    -e "SITE_NAME=$SITE_NAME" \
    -e "DB_ROOT_USER=$DB_ROOT_USER" \
    -e "MYSQL_ROOT_PASSWORD=$MYSQL_ROOT_PASSWORD" \
    -e "ADMIN_PASSWORD=$ADMIN_PASSWORD" \
    -e "INSTALL_APPS=erpnext" \
    mysitecom_erpnext-python_1 docker-entrypoint.sh new

Attempt 1 to connect to mariadb:3306
Attempt 1 to connect to redis-queue:6379
Attempt 1 to connect to redis-cache:6379
Attempt 1 to connect to redis-socketio:6379
Connections OK

Traceback (most recent call last):
  File "/home/frappe/frappe-bench/commands/new.py", line 118, in <module>
    main()
  File "/home/frappe/frappe-bench/commands/new.py", line 70, in main
    db_port=db_port,
  File "/home/frappe/frappe-bench/apps/frappe/frappe/installer.py", line 72, in _new_site
[...]

PR #507, opened 2021-08-05
fix database access errors for new builds

8633brown:
This adds the `no_mariadb_socket` flag for the `site-creator` service.
By adding this flag the frappe framework should now add the sql user using the `%` wildcard to allow the generated frappe user to connect over the docker network fixing #448.


8633brown:
My bad doesn't seem to fix it after checking again.

PR #509, opened 2021-08-10
Make sure apt lists cleaned up in Dockerfile

PeterDaveHello:
Some of the Dockerfiles missed the part to clean up apt lists, this PR make sure every Dockerfile to do it!

PR #510, opened 2021-08-10
Clean up apt cache in dockerfile

revant:
#509


PeterDaveHello:
@revant Can we just do it in #509?

PR #513, opened 2021-08-30
Update README.md

ittyabrehman:
Installation video tutorial added

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@
 
 Wait for 5 minutes for ERPNext site to be created or check `site-creator` container logs before opening browser on port 80. (username: `Administrator`, password: `admin`)
 
+### Video tutorial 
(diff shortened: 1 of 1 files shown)


Thunderbottom:
Hey, thanks for the pull request. Although, I don't really think a video is needed for anyone to run a docker setup since I think it's fairly straightforward if one understands what docker is. The video is also quite long and includes stuff specific to Windows, kind of defeats the purpose of adding it to the repository given that the docker image can be run multiple platforms.

I'll be closing this pull request for now, but if this gains enough interest maybe we can create a short asciinema demonstrating the docker setup and add that to the repository instead.

revant:
I went through the video.

- it's a call recording
- most of it is docker installation on windows.
- at the end there is no "bench start". Or any development. The bench init gets stuck 2 times and video ends as the call ends.

Add a wiki page if you wish https://github.com/frappe/frappe_docker/wiki

This is a high quality video for production installation. https://youtu.be/_08L9pubPCY

PR #533, opened 2021-10-30
ci: fixes ci steps

revant:
helm release deploy key uses webfactory/ssh-agent@v0.5.3
test only on PR
docker login only if not PR

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -88,6 +88,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: docker/setup-buildx-action@v1
       - uses: docker/login-action@v1
+        if: github.event_name != 'pull_request'
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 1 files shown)


revant:
@vrslev

vrslev:
> docker login only if not PR
Yep, it should work this way. Also, probably shouldn't login and push if workflow is running on fork. I will dig into this.

> again it didn't trigger tests?
That's weird. Look into this also

vrslev:
Can you merge this so my changes don't overwrite code relevant to helm chart?

vrslev:
There's too much stuff in single workflow. Probably it is a good idea to at least split up dev and stable builds. What do you think?

vrslev:
> again it didn't trigger tests?

That is weirder because PR #531 triggered the workflow

revant:
we'll separate the workflow a needed. you can send a PR?

vrslev:
Sure

vrslev:
This is why workflow wasn't triggered, syntax error:

Screenshot 2021-10-30 at 13 58 56

I'll fix in new PR.

Issue #526, opened 2021-10-27
Missing packages required for WeasyPrint in frappe-worker 


dtwm:
## Description of the issue
Image for `frappe-worker` doesn't have some packages required by `weasyprint` package. Due to that when trying to start ERP-next `site-creator` fails.

## Context information
While using `develop` branch `site-creator` failed on importing `weasyprint` with message of missing `libcairo-2` package

## Steps to reproduce the issue
0. Checkout develop branch
1. Copy `env-local` to `.env`
2. Start project with docker-compose
3. Logs of `site-creator` shows critical error on importing `weasyprint`
4. Project isn't configured and doesn't work.

## Solution
Add more packages in frappe-worker Dockerfile. 
I had to add `libcairo2 python3-cffi python3-brotli libpango-1.0-0 libpangoft2-1.0-0 libpangocairo-1.0-0` but I'm not really familiar with weasyptint so that might be excessive. 
Weasyprint documentation state https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#debian-11 that for debian only `python3-cffi python3-brotli libpango-1.0-0 libpangoft2-1.0-0` are needed but I was not able to run without explicit installation of `libpangocairo`.

vrslev:
@dtwm Will you send PR to fix this?

dtwm:
sorry, was away for a while.

todfox:
Same issue. Anything I can do to help?

revant:
merged #537

PR #537, opened 2021-11-01
fix: Add missing packages for frappe:develop

dtwm:
PR to address issue #526
In develop version of frappe WeasyPrint is used that requires few libs on host system. 
Suggested changes:
+ install required packages if image is build for developer branch

Notes:
+ install only for developer branch. If build is for branch based on develop, nothing will happen. This might be solved by some additional agruments
+ list of requirements is taken from packages documentation. Altho I was not able to reproduce that `python3-cffi` `python3-brotli` are needed 100%.


build/frappe-worker/Dockerfile
```diff
@@ -40,6 +40,16 @@ RUN apt-get update -y && apt-get install \
     # For arm64 python wheel builds
     gcc \
     g++ -y \
+    # Install additional requirements for develop branch
+    && if [ "${GIT_BRANCH}" = 'develop' ]; then \
+        apt install -y \
```
vrslev, line 45: Use apt-get

dtwm: right.

revant (approved):
LGTM

todfox:
pushed to hub.docker.com?

vrslev:
Develop images update every day at 12 am.

PR #544, opened 2021-11-05
Fix builds and tests

vrslev:
- Build and push Frappe and ERPNext images separately to make sure ERPNext images uses latest Frappe images and fix opacity images in tests
- Add simple and fast tests that ping site for Frappe and ERPNext images before pushing
- Refactor integration test and fix for GitHub Actions
- Add integration test workflow (runs every day at 1:00 am)
- Add test for bench image
- Add lint workflow to check shell style


vrslev:
@revant Merging?

revant:
lint failed?

revant (changes requested):
current lint rules should not fail. can it be configured?

revant:
this is also fixed in PR?
https://github.com/frappe/frappe_docker/runs/4123183231?check_suite_focus=true

vrslev:
> current lint rules should not fail.

I would better keep strict configuration and fix all of these in the next PR. What you say?

vrslev:
> this is also fixed in PR?

Yes

revant:
> I would better keep strict configuration and fix all of these in the next PR. What you say?

merged this

vrslev:
Temporarily disabled lint worklow

PR #553, opened 2021-11-08
Fix lint issues

vrslev:
This fixes possible issues that came up while using vanilla shellcheck (with -x option only)


vrslev:
@revant There's one issue left. Do we have to use `ls` and not `find`?
https://github.com/frappe/frappe_docker/blob/d1ed880c7825da06d07522b849ee03bd5bbd29e3/build/frappe-nginx/docker-entrypoint.sh#L11

vrslev:
Don't merge. Breaks ping test after build.

vrslev:
All good now.

revant:
> @revant There's one issue left. Do we have to use `ls` and not `find`?

this `touch` command came as a PR, it can be improved.

vrslev:
What would be equivalent for find?

revant:
> What would be equivalent for find?

how about this?

```
find /assets -maxdepth 1 -name "*.*" | head -n 1
```

vrslev:
Seems to work fine. I think we can merge after CI passes

vrslev:
This PR also fixes integration test: https://github.com/frappe/frappe_docker/runs/4147020329?check_suite_focus=true

PR #561, opened 2021-11-10
ci: fix git creds for helm release

revant:
- Removes tests for stable. In case of release trigger we don't want anything to fail.
- add git creds action

### DIFF
--- a/.github/workflows/build_stable.yml
+++ b/.github/workflows/build_stable.yml
@@ -68,9 +68,6 @@ jobs:
           targets: frappe-stable-test
           load: true
 
-      - name: Test
-        run: ./tests/test-frappe.sh
-
       - name: Push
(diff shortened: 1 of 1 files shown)


vrslev:
@revant I think it is bad thing to remove test before stable build. It's better not to release failed build that do.

revant:
> @revant I think it is bad thing to remove test before stable build. It's better not to release failed build that do.

Frappe/ERPNext release has already happened. People need to upgrade it and try to check for docker tags for the version.
If there is no release on docker the main release really doesn't stop.

vrslev:
No, I get it. I'm talking about disabling tests. Would you rather release broken Docker images or let people wait a little to fix it (if they're actually broken)?

revant:
https://github.com/frappe/frappe_docker/issues/565

PR #569, opened 2021-11-18
Fix permission denied error

Meradoch:
While building the frappe-worker image, a permissions denied error was being thrown while executing line 80:
`RUN mkdir -p apps logs commands sites /home/frappe/backups`

I now first create the folder as root, change ownership to user frappe before changing to user frappe.

### DIFF
--- a/build/frappe-worker/Dockerfile
+++ b/build/frappe-worker/Dockerfile
@@ -67,16 +67,12 @@ RUN if [ "$(uname -m)" = "aarch64" ]; then export ARCH=arm64; fi \
     && dpkg -i $downloaded_file \
     && rm $downloaded_file
 
-RUN chown -R frappe:frappe /home/frappe
-
 # Setup docker-entrypoint
 COPY build/frappe-worker/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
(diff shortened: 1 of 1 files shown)


build/frappe-worker/Dockerfile
```diff
@@ -73,9 +73,12 @@ RUN chown -R frappe:frappe /home/frappe
 COPY build/frappe-worker/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
 RUN ln -s /usr/local/bin/docker-entrypoint.sh / # backwards compat
 
-USER frappe
 WORKDIR /home/frappe/frappe-bench
 
+RUN chown -R frappe:frappe /home/frappe/frappe-bench
```
revant, line 78: This like does same for the parent directory.

https://github.com/frappe/frappe_docker/blob/ccc3fc27073b1c6ccd983745d22ef3dc07308aea/build/frappe-worker/Dockerfile#L70

I think this line needs to be moved to line 78 instead of the line you added?

revant (changes requested):

Meradoch (commented):
This seems cleaner, and the error is gone the build completes successfully

Issue #567, opened 2021-11-11
Permission denied for nginx container after enable SELinux in docker daemon 


Tus1688:
## Description of the issue
Hi everyone! Thank you for making ERP next deployment become easier and less complicated. I tried to harden docker daemon using SELinux. Luckily, this happen to my testing server, I create  /etc/docker/daemon.json then fill it with 
```
{
        "selinux-enabled": true
}
```
FYI, I have enabled selinux before updating to latest version of ERP next, and it worked as it supposed to be. Then.... everything has change after rebuilding container to latest version using `docker-compose pull` and `docker-compose up -d`, I got `bad gateway` when trying to access our sites. 

OS: Rocky Linux 8.4
Docker: Docker version 20.10.10, build b485636
Compose: docker-compose version 1.29.2, build 5becea4c

## Steps to reproduce the issue

1. Create /etc/docker/daemon.json
2. Fill it with 
```
{
        "selinux-enabled": true
}
```
3. `systemctl restart docker`
4. `restorecon -R -v /var/lib/docker`
4. `restorecon -R -v /usr/bin`
5.  `cd /frappe-docker`
6.  `docker-compose pull`
7.  `docker-compose up -d`
8.  Accessing out sites 
9.  I got `Bad Gateway` in top left corner
10.  `docker ps`
11.  Everything is running normally except `frappe/erpnext-nginx:version-13` *logs below

### Observed result
Get `Bad Gateway`
### Expected result
Can access our site normally
### Stacktrace / full error message if available

```
rsync: [receiver] mkstemp "/assets/frappe/node_modules/@snyk/inquirer/.package.json.UaHnNY" failed: Permission denied (13)
rsync: [receiver] mkstemp "/assets/frappe/node_modules/@snyk/inquirer/lib/.inquirer.js.tXEiEY" failed: Permission denied (13)
rsync: [receiver] mkstemp "/assets/frappe/node_modules/@snyk/inquirer/lib/objects/.choice.js.Akuf6X" failed: Permission denied (13)
rsync: [receiver] mkstemp "/assets/frappe/node_modules/@snyk/inquirer/lib/objects/.choices.js.L0oHlX" failed: Permission denied (13)
rsync: [receiver] mkstemp "/assets/frappe/node_modules/@snyk/inquirer/lib/objects/.separator.js.usVRjY" failed: Permission denied (13)
[...]

PR #570, opened 2021-11-18
feat(frappe-nginx): use nginxinc/nginx-unprivileged image

revant:
#567

### DIFF
--- a/build/frappe-nginx/Dockerfile
+++ b/build/frappe-nginx/Dockerfile
@@ -44,12 +44,16 @@ RUN git clone --depth 1 https://github.com/frappe/bench /tmp/bench \
 RUN cp -R apps/frappe/frappe/public/* sites/assets/frappe \
     && cp -R apps/frappe/node_modules sites/assets/frappe/
 
-FROM nginx:latest
+FROM nginxinc/nginx-unprivileged:latest
 
 COPY --from=builder /home/frappe/frappe-bench/sites /var/www/html/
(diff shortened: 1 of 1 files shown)


Thunderbottom (approved):
LGTM! builds fine.

Tus1688:
After upgrading to the latest version of frappe docker. Docker Daemon that hardened with selinux still causing issue with nginx image. Any clue?

Thunderbottom:
Could you try adding SELinux rules for the folder that you're trying to mount within the container?

Tus1688:
Hi @Thunderbottom, I am not sure about it. All I do is changing `/etc/docker/daemon.json`  by adding `selinux-enabled: true`

Thunderbottom:
Well, right now I don't really have a system that runs selinux, and neither do I have the capacity for the mental gymnastics that one requires to set it up correctly. Maybe for now you can try replacing `:rw` with the `:z` option in your docker-compose's volume mountpoint and report back:

```yaml
volumes:
  - sites-vol:/var/www/html/sites:z
  - assets-vol:/assets:z
```

The `:z` option should basically ask selinux to relabel the filesystem correctly so that it can be mounted inside the container. Let me know how it goes for you.

Tus1688:
I have tried them and unfortunately it's not working, I also did `restorecon -R -v /var/lib/docker` and `restorecon -R -v /usr/bin`

Issue #574, opened 2021-11-21
ERPNext Docker Edge Latest pull failing to start


yabdali:
Hi,
I had a running docker instance of ERPNext and decided to stop it, pull the latest edge image to update.  Unfortunately, the latest edge release broke my setup, I keep getting errors after starting the container and I am not sure where those errors are coming from.  Appreciate your guidance and support.

erpnext-nginx_1           | rsync: [generator] chgrp "/assets/frappe/node_modules/read-pkg/node_modules/path-type" failed: Operation not permitted (1)
erpnext-nginx_1           | rsync: [generator] chgrp "/assets/frappe/node_modules/readable-stream" failed: Operation not permitted (1)
erpnext-nginx_1           | rsync: [generator] chgrp "/assets/frappe/node_modules/readable-stream/doc" failed: Operation not permitted (1)
erpnext-nginx_1           | rsync: [generator] chgrp "/assets/frappe/node_modules/readable-stream/doc/wg-meetings" failed: Operation not permitted (1)
erpnext-nginx_1           | rsync: [generator] chgrp "/assets/frappe/node_modules/readable-stream/lib" failed: Operation not permitted (1)
erpnext-nginx_1           | rsync: [generator] chgrp "/assets/frappe/node_modules/readable-stream/lib/internal" failed: Operation not permitted (1)
erpnext-nginx_1           | rsync: [generator] chgrp "/assets/frappe/node_modules/readable-stream/lib/internal/streams" failed: Operation not permitted (1)
erpnext-nginx_1           | rsync: [generator] chgrp "/assets/frappe/node_modules/readable-stream/node_modules" failed: Operation not permitted (1)
erpnext-nginx_1           | rsync: [generator] chgrp "/assets/frappe/node_modules/readable-stream/node_modules/isarray" failed: Operation not permitted (1)
erpnext-nginx_1           | rsync: [generator] chgrp "/assets/frappe/node_modules/redent" failed: Operation not permitted (1)
erpnext-nginx_1           | rsync: [generator] chgrp "/assets/frappe/node_modules/redent/node_modules" failed: Operation not permitted (1)
erpnext-nginx_1           | rsync: [generator] chgrp "/assets/frappe/node_modules/redent/node_modules/.bin" failed: Operation not permitted (1)
erpnext-nginx_1           | rsync: [generator] chgrp "/assets/frappe/node_modules/redent/node_modules/.bin/strip-indent" failed: Operation not permitted (1)
[...]

PR #576, opened 2021-11-21
fix: change unpriviliged nginx port from 80 to 8080

revant:
fixes https://github.com/frappe/frappe_docker/issues/574#issuecomment-974774363

```
2021/11/21 08:10:38 [emerg] 1#1: bind() to 0.0.0.0:80 failed (13: Permission denied)
```


revant:
assets-vol, sites-vol, logs-vol need to be 1000:1000 uid:gid permission before starting nginx. Any ideas?

vrslev:
Do we need to use `nginxinc/docker-nginx-unprivileged` image?

revant:
> Do we need to use `nginxinc/docker-nginx-unprivileged` image?

mostly yes.
it is ready made image that allows non root user to run nginx.
it does need some tweaks to configs.

https://github.com/nginxinc/docker-nginx-unprivileged/blob/02a4307966e5b5d07226cb7251afa821adb0e297/Dockerfile-debian.template#L100

vrslev:
> it is official image by nginx inc.

ok, didn’t know that :)

PR #580, opened 2021-11-22
Fix nginx bind to port 80 error

Meradoch:
This fixes `nginx: [emerg] bind() to 0.0.0.0:80 failed (13: Permission denied)` error as only root user can bind to port 80.


revant:
we don't want to use root user! All the containers now are non-root.
The port is now changed to 8080 from 80 so non root nginx can work.

pull the latest stable or edge image it will publish port 8080

Meradoch:
Seen it. Thanks!

PR #581, opened 2021-11-23
Removed a Tab Space for Traefik Labels

RohithCIS:
Traefik doesn't detect if the labels are indented another level 

I was stuck for over an hour trying to figure out if I setup something wrong. I shouldn't have copy pasted without checking once.

Fixing Doc so other shouldn't have to go through this.

### DIFF
--- a/docs/docker-swarm.md
+++ b/docs/docker-swarm.md
@@ -156,18 +156,18 @@ services:
     deploy:
       restart_policy:
         condition: on-failure
-      labels:
-        - "traefik.docker.network=traefik-public"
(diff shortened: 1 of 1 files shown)


revant:
of label should be nested in case of using docker stack deploy command. 

https://docs.docker.com/compose/compose-file/compose-file-v3/#deploy check docs here.

I've been using it nested under deploy in production without any error.

RohithCIS:
I'm using portainer over a swarm cluster. Indentation did not work for me. I had to untab a level for traefik to detect the router config from the labels.

PR #586, opened 2021-11-26
ci: Don't rebuild develop ERPNext images on Push step

vrslev:
Now it rebuilds erpnext-worker: https://github.com/frappe/frappe_docker/runs/4329208058?check_suite_focus=true  🧐

### DIFF
--- a/.github/workflows/build_develop.yml
+++ b/.github/workflows/build_develop.yml
@@ -119,5 +119,5 @@ jobs:
         uses: docker/bake-action@v1.6.0
         with:
           files: docker-bake.hcl
-          targets: frappe-develop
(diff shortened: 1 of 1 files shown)


vrslev:
Figured it out!

`erpnext-nginx` doesn't build again because it caches intermediate image based on `node` and just copies assets on `frappe/frappe-nginx` or `localhost:5000/frappe/frappe-nginx`—it takes very little time. But `erpnext-worker` is based on `frappe/frappe-worker` and it is replaced with `localhost:5000/frappe/frappe-worker` in tests.

So, there's nothing we can do about this rebuild, I think.

PR #597, opened 2021-12-05
Upgrade mariadb

vrslev:
Upgrade mariadb from 10.3 to 10.6, use `command` parameter in docker-compose files instead of copying mysql config file.
This works fine for me in production for a while.

@revant Couple of questions.
1. How do you feel about inline db configuration?
2. Will this break Try in Play With Docker config?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -139,12 +139,19 @@ services:
       - redis-socketio-vol:/data
 
   mariadb:
-    image: mariadb:10.3
+    image: mariadb:10.6
     restart: on-failure
+    command:
+      - --character-set-server=utf8mb4
+      - --collation-server=utf8mb4_unicode_ci
+      - --skip-character-set-client-handshake
+      - --skip-innodb-read-only-compressed # Temporary fix for MariaDB 10.6
     environment:
       - MYSQL_ROOT_PASSWORD=${MYSQL_ROOT_PASSWORD}
+      # Sometimes db initialization takes longer than 10 seconds and site-creator goes away.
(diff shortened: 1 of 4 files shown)


revant:
> 1. How do you feel about inline db configuration?

I don't mind to pass cli args. Will it configure everything as needed here https://github.com/frappe/bench/wiki/MariaDB-conf-for-Frappe? Or we don't need to have advance configuration for simple docker setups?

Also consider this https://discuss.erpnext.com/t/table-tabdefaultvalue-missing-error-when-creating-new-site/78019/10 if using 10.6

> 2. Will this break Try in Play With Docker config?

you can try it out. You just need to give it a yml from any where, try from custom gist

e.g. `https://labs.play-with-docker.com/?stack=https://raw.githubusercontent.com/frappe/frappe_docker/develop/tests/pwd.yml` instead of `https://raw.githubusercontent.com/frappe/frappe_docker/develop/tests/pwd.yml` try your custom raw gist file.

revant (approved):

vrslev:
> Will it configure everything as needed here

Yes! The db works exactly as it should be. I've been using this setup for several months.

> Also consider this https://discuss.erpnext.com/t/table-tabdefaultvalue-missing-error-when-creating-new-site/78019/10 if using 10.6

`--skip-innodb-read-only-compressed` flag fixes this.

> you can try it out
> Also, if you edit the pwd file, remove the traces of develop branch from README link to PWD. 😅

Will do. I'll send another PR as you merged this one already.

PR #605, opened 2021-12-06
fix: update devcontainer mariadb

revant:


revant:
@vrslev merge if everything good.

PR #650, opened 2022-02-03
Change mysql.user table to mysql.global_priv table for MariaDB 10.4+

Meradoch:
Since frappe_docker now uses the mariadb:10.6 docker image for its database, the `UPDATE mysql.user SET Host = '%' where User = 'db_name'; FLUSH PRIVILEGES;` command fails as _mysql.user_ is deprecated and is now a view to the _mysql.global_priv_ table as from MariaDB 10.4+: [https://mariadb.com/kb/en/mysqluser-table/](url).


vrslev:
Failing lint in CI is irrelevant

PR #617, opened 2021-12-17
Global refactoring

vrslev:
# Summary

What was done:
- Clean up, unify and _uncomplicate_ the code: some things were removed, some merged together.
- Refactor integration tests: now they're written in Python, more efficient, precise and meticulous 😃
- Refactor DevOps utils: much faster, less flaky, easier to maintain and use.

You can check out changes yourself locally. To do this, clone [my fork](https://github.com/vrslev/frappe_docker/tree/beautify), checkout to `beautify` branch. Then build images with `FRAPPE_VERSION=... ERPNEXT_VERSION=... docker buildx bake`. Next step is to start everything up, here are docs: https://github.com/vrslev/frappe_docker/blob/beautify/docs/images-and-compose-files.md, https://github.com/vrslev/frappe_docker/blob/beautify/docs/setup-options.md.

After the PR is merged, your current setup _will_ break. There're no migration guide, but we (@revant and I) updated [the docs](https://github.com/vrslev/frappe_docker/blob/beautify/README.md), so it shouldn't be a problem. If you experience difficulties with upgrading, feel free to leave an issue.

---
# Why?

Currently there's a lot of stuff that could be done better. I'm trying to improve performance, logic behind processes and maintainer happiness.

# What?

Almost everything.

## CI

- Move `get-latest-tags.sh` to Python (`get_latest_tags.py`). Kinda preparation for #616.
- Add `docker-build-push.yml` reusable workflow and use it in `build_develop.yml` and `build_stable.yml`. It is clearer this way.
- Remove `test.yml` and add integration test to build workflows.
- Correct `dependabot.yml` according to changes.

## Images

- Rename target names in frappe/bench.
- Remove frappe/erpnext-worker and frappe/erpnext-nginx separate Dockerfiles. They are now embedded in frappe/frappe-worker and frappe/frappe-nginx Dockerfiles using multi-stage builds. It is much more efficient this way.
- Move nginx Dockerfile from frappe-nginx to nginx and worker Dockerfile from frappe-worker to worker.
- Don't include development node modules in final nginx images.
- Reuse stages for efficient builds in nginx images.
- Update Nginx config from frappe/bench upstream.
- Make Nginx config more opinionated. Who wants can mount their own.
- Simplify frappe/frappe-socketio build.
- Remove majority of frappe/frappe-worker commands since there's pretend bench (auto_migrate, backup, check_connection, constants, doctor, gevent_patch, migrate, new, restore_backup and utils modules are gone).
[...]

PR #669, opened 2022-02-23
fix: check if s3 object is a directory by checking the start of the object `ContentType` property

crosvera:

After uploading the backups. and checking for the local files to download, `push_backup.py` specifically `delete_old_backups` method assumes that S3 objects that are actually directories will be only marked as `application/x-directory` in the `ContentType` object property. But there are some cases -like mine- that has more information such us: `application/x-directory; charset=UTF-8`.

This fix only changes the `ContentType` check by using `startswith()`

### DIFF
--- a/build/frappe-worker/commands/push_backup.py
+++ b/build/frappe-worker/commands/push_backup.py
@@ -75,7 +75,7 @@ def delete_old_backups(limit, bucket, site_name):
         for obj in objects.get("CommonPrefixes"):
             if obj.get("Prefix") == bucket_dir + "/":
                 for backup_obj in bucket.objects.filter(Prefix=obj.get("Prefix")):
-                    if backup_obj.get()["ContentType"] == "application/x-directory":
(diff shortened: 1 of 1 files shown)


revant:
You may have to send this again after refactor is merged.

vrslev:
Yep, @crosvera follow #617

vrslev:
Hey there! We don't delete old backups after refactor for the same of simplicity. If you want to do that, use custom scripts.

revant:
push backups are tricky.

when the size of data grows different strategies need to be used. simple dump and tar becomes inefficient.

there is new push-backup script you can check.

ideally for files use restic as it'll do snapshots and backups in increments.

for db use volume snapshot as it is faster and efficient than overloading processor and memory for taking db backups.

vrslev:
@revant good to know, actually. Closing?

revant:
I'll close.

@crosvera check the new script. If you wish to improve it, send update to new script.

PR #676, opened 2022-03-07
fix: make changes for backward compat

revant:
make changes for refactor to work on existing setups


vrslev:
Why?

vrslev:
If we're merging that one, this one will be gone. Maybe create another branch?

revant:
I'll create another branch.

Just for reference. If anyone doesn't want to setup the new compose then they can just pull this "branch" and keep working with existing setup.

What should the branch be called?

revant:
https://github.com/frappe/frappe_docker/tree/compat

PR #677, opened 2022-03-08
fix: new site command on Frappe >13.21.0

vrslev:
After https://github.com/frappe/frappe/pull/16134
`TypeError: _new_site() got an unexpected keyword argument 'mariadb_root_username'`


vrslev:
Integration test is irrelevant.

PR #694, opened 2022-03-14
Fix configurator command on some platforms

vrslev:
Turns out _on some machines_ `configurator` serivce fails because of path resolving (I guess that's why). I encountered this error when debugging custom app Docker setup in GitHub Actions. Also https://github.com/vrslev/frappe_docker/pull/26


vrslev:
False alarm: I accidentally used old images 😅

Issue #695, opened 2022-03-14
Migration fails due to theme


revant:
```
Traceback (most recent call last):
  File "/usr/local/bin/patched_bench_helper.py", line 46, in <module>
    raise SystemExit(main())
  File "/usr/local/bin/patched_bench_helper.py", line 41, in main
    frappe.utils.bench_helper.main()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/utils/bench_helper.py", line 19, in main
    click.Group(commands=commands)(prog_name='bench')
  File "/home/frappe/frappe-bench/env/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/frappe/frappe-bench/env/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/frappe/frappe-bench/env/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/frappe/frappe-bench/env/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/frappe/frappe-bench/env/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/frappe/frappe-bench/env/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/frappe/frappe-bench/env/lib/python3.9/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/commands/__init__.py", line 27, in _func
    ret = f(frappe._dict(ctx.obj), *args, **kwargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/commands/site.py", line 309, in migrate
    migrate(
  File "/home/frappe/frappe-bench/apps/frappe/frappe/migrate.py", line 91, in migrate
    frappe.get_attr(fn)()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/website/doctype/website_theme/website_theme.py", line 187, in after_migrate
    doc.generate_bootstrap_theme()
[...]

PR #700, opened 2022-03-14
fix: add entrypoint to worker

revant:
symlinks frappe node_modules from sites/assets to apps/frappe

#695

### DIFF
--- a/images/worker/Dockerfile
+++ b/images/worker/Dockerfile
@@ -114,13 +114,13 @@ CMD [ "/home/frappe/frappe-bench/env/bin/gunicorn", "-b", "0.0.0.0:8000", "frapp
 FROM configured_base as frappe
 
 RUN echo "frappe" >/home/frappe/frappe-bench/sites/apps.txt
-COPY --from=frappe_builder /home/frappe/frappe-bench/apps/frappe /home/frappe/frappe-bench/apps/frappe
-COPY --from=frappe_builder /home/frappe/frappe-bench/env /home/frappe/frappe-bench/env
+COPY --from=frappe_builder --chown=frappe:frappe /home/frappe/frappe-bench/apps/frappe /home/frappe/frappe-bench/apps/frappe
+COPY --from=frappe_builder --chown=frappe:frappe /home/frappe/frappe-bench/env /home/frappe/frappe-bench/env
 
 
 FROM configured_base as erpnext
 
 RUN echo "frappe\nerpnext" >/home/frappe/frappe-bench/sites/apps.txt
-COPY --from=frappe_builder /home/frappe/frappe-bench/apps/frappe /home/frappe/frappe-bench/apps/frappe
-COPY --from=erpnext_builder /home/frappe/frappe-bench/apps/erpnext /home/frappe/frappe-bench/apps/erpnext
-COPY --from=erpnext_builder /home/frappe/frappe-bench/env /home/frappe/frappe-bench/env
+COPY --from=frappe_builder --chown=frappe:frappe /home/frappe/frappe-bench/apps/frappe /home/frappe/frappe-bench/apps/frappe
+COPY --from=erpnext_builder --chown=frappe:frappe /home/frappe/frappe-bench/apps/erpnext /home/frappe/frappe-bench/apps/erpnext
+COPY --from=erpnext_builder --chown=frappe:frappe /home/frappe/frappe-bench/env /home/frappe/frappe-bench/env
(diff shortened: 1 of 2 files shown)


revant:
```
ln: failed to create symbolic link '/home/frappe/frappe-bench/apps/frappe/node_modules': Permission denied
```

Any suggestion how to solve this?

images/worker/Dockerfile
```diff
@@ -102,22 +102,25 @@ COPY pretend-bench.sh /usr/local/bin/bench
...
+
 CMD [ "/home/frappe/frappe-bench/env/bin/gunicorn", "-b", "0.0.0.0:8000", "frappe.app:application" ]
 
 
 FROM configured_base as frappe
 
 RUN echo "frappe" >/home/frappe/frappe-bench/sites/apps.txt
-COPY --from=frappe_builder /home/frappe/frappe-bench/apps/frappe /home/frappe/frappe-bench/apps/frappe
-COPY --from=frappe_builder /home/frappe/frappe-bench/env /home/frappe/frappe-bench/env
+COPY --from=frappe_builder --chown=frappe:frappe /home/frappe/frappe-bench/apps/frappe /home/frappe/frappe-bench/apps/frappe
```
vrslev, line 117: Better not to allow non-root users to change code. Security.

images/worker/Dockerfile
```diff
@@ -102,22 +102,25 @@ COPY pretend-bench.sh /usr/local/bin/bench
 COPY push_backup.py /usr/local/bin/push-backup
 COPY configure.py patched_bench_helper.py /usr/local/bin/
 COPY gevent_patch.py /opt/patches/
+COPY entrypoint.sh /usr/local/bin/entrypoint.sh
 
 WORKDIR /home/frappe/frappe-bench/sites
 
+ENTRYPOINT [ "entrypoint.sh" ]
```
vrslev, line 109: Use absolute path

images/worker/entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+# Link Frappe's node_modules/ to make Website Theme work
+if test -d /home/frappe/frappe-bench/sites/assets/frappe/node_modules; then
```
vrslev, line 6: What if we make broken symlink in Dockerfile instead and when assets volume is there it would be valid

vrslev (changes requested):

images/worker/Dockerfile
revant: this causes permission denied for `ln` command

images/worker/entrypoint.sh
revant: does it work as expected?

vrslev: Don’t know. Can you test?

images/worker/Dockerfile
revant: I'll prefer frappe and other apps be installed by user frappe and not root. This was what was done in previous image.

It may be anti-pattern but it make patching easy. For depending images as well as runtime.

images/worker/entrypoint.sh
revant: okay.

images/worker/Dockerfile
vrslev: > It may be anti-pattern but it make patching easy.

In images based on Frappe it is easy to switch to root and back to frappe user. If you think it is the best to give rights to frappe user, let it be so.

images/worker/entrypoint.sh
revant: this worked. I reset the changes in PR, no `--chown` and `entrypoint.sh` used. Then I made following changes to frappe_builder

```
FROM build_deps as frappe_builder

ARG FRAPPE_VERSION
RUN --mount=type=cache,target=/home/frappe/.cache/pip \
    git clone --depth 1 -b ${FRAPPE_VERSION} https://github.com/frappe/frappe apps/frappe \
    && env/bin/pip install -e apps/frappe \
    && env/bin/pip install -U gevent \
    && rm -r apps/frappe/.git \
    && mkdir -p /home/frappe/frappe-bench/sites/assets/frappe/node_modules \
    && ln -s /home/frappe/frappe-bench/sites/assets/frappe/node_modules /home/frappe/frappe-bench/apps/frappe/node_modules
```

I had to create a directory and make a symlink. That became active in container. I could save website theme.

vrslev: That is great!

revant: I'll close this PR and create a fresh one with only this change.

PR #702, opened 2022-03-15
fix: link frappe node_modules to make Website Theme work

revant:


vrslev:
Everything works, right? Do you think we should add test for this?

revant:
I tested it on simple compose as well as in the helm chart/kuberentes. It is working.

I followed these steps:

Install > Create Site > Run Wizard > Save Website Theme

What do you plan to test? We can skip the wizard part and directly use orm : `frappe.new_doc('Website Theme')`

There will be many such things we may find. Should we test all these cases? also why test bad design? We reinforce the bad design with tests?

vrslev:
The point of automating testing is to ensure that with time or new changes code still works as it supposed to — doesn't matter if code is around bad design or not. Although if added thing is trivial and obvious then tests are optional.

In this case it makes sense to add _integration_ test (not unit test) — that is kinda what you suggested. If you have troubles with understanding test suite that I brought in, I'm happy to help 😃

revant:
how to test?

Like this?

```shell
curl -X POST http://mysite.localhost:8080/api/method/login --cookie-jar /tmp/cookie.txt \                                                                
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json' \
  -d '{"usr":"Administrator","pwd":"admin"}'

curl --cookie /tmp/cookie.txt -X POST -H 'Accept: application/json' -d '{"theme":"node_modules_not_found"}' http://mysite.localhost:8080/api/resource/Website%20Theme
```

vrslev:
I would just add a test in [`main.py`](https://github.com/frappe/frappe_docker/blob/main/tests/main.py). I can do it if you want

revant:
> I would just add a test in [`main.py`](https://github.com/frappe/frappe_docker/blob/main/tests/main.py). I can do it if you want

Yes! please co-commit?

Should this be the approach? Login with Admin and test Theme creation?

vrslev:
I added the test. Before it would fail (I checked that), not it doesn't.

revant:
tests failed? they suppose to work before merge?

vrslev:
They didn't suppose to fail. Fixing it :)

revant:
merge?

vrslev (approved):

vrslev:
Yes!

PR #705, opened 2022-03-16
Use pytest

vrslev:
Use `pytest` test runner for our integrations tests. Features that matter:
- run specific test (not whole suite),
- fixtures and DI for code reusability.


vrslev:
Test are supposed to fail. See result here: https://github.com/vrslev/frappe_docker/actions/runs/1997187255, https://github.com/vrslev/frappe_docker/actions/runs/1997186938

vrslev:
@revant Can you check this out?

Issue #645, opened 2022-01-31
Changing the UID/GID from 1000 to something different without rebuilding the images


mwllgr:
## Description of the issue
The 1000 user is used for another purpose on my system (ISPconfig), when setting the volume directory to 1000, I essentially allow unwanted guests in these folders.

### Observed result
Web clients (they are somehow connected with my user 1000) might be able to access my ERP system.

### Expected result
One should be able to change the used UID/GID.  Is this possible somehow?

revant:
right now only way is to build image. 
even during build it runs chown.

PR #714, opened 2022-03-19
Simplify Dockerfiles and custom app guide

### LINKED ISSUE #697: Release assets builder along with frontend image
The idea is to ship pre-built building blocks for custom apps. This will take a little effort and will save a lot of build time (and will help to save the planet). Also it would be clearer for users how to built their own images.

Partially resolves issue with custom apps based on ERPNext: https://github.com/frappe/frappe_docker/pull/617#issuecomment-1065440200

vrslev:
This PR adds new Docker image `assets-builder`. It uncomplicates custom app building. Before, we released only pre-built frontend assets. Now we also provide builder.

Closes #697.


revant:
it can build multiple apps?

if it solves 80% of custom builds it's okay to merge.

but voice and issues will come from the 20% with custom egde cases. The 80% will be silent. The resulting effect is, maintainers feel feature is incomplete.

That's what happened with install_app script used before refactor. Contributor had noble idea. Later there was total neglect to the script and initial contributor went missing.

vrslev:
@revant I have an idea how to solve issue with multiple custom apps and still keep it simple, robust.

The idea is to make a script that does what is done in complicated multistage build. I will work on it and update this PR.

This is fun!

vrslev:
@revant Just added new commits. Do they look legit to you?

custom_app/backend.Dockerfile
```diff
@@ -1,8 +1,20 @@
...
+
+# or with git:
+# ARG APP_NAME
+# ARG BRANCH
+# ARG GIT_URL
+# RUN install-assets ${APP_NAME} ${BRANCH} ${GIT_URL}
 
-RUN echo "frappe\n${APP_NAME}" >/home/frappe/frappe-bench/sites/apps.txt \
-    && ../env/bin/pip install --no-cache-dir -e ../apps/${APP_NAME}
+USER frappe
```
revant, line 20: Can uid/gid be `--build-arg` defaulting to 1000/frappe?

People who really need different gid/uid can override it in their custom apps then.

https://github.com/frappe/frappe_docker/issues/645

vrslev: It would be quite confusing if users would rebuild images themselves with different uid and name them the same way official images are named. In this case I would suggest to add another Dockerfile and change user there.

revant: Part of docs I guess. I'll resolve.

revant:
frappe/assets-builder registry needed on `hub.docker.com`?

vrslev:
@revant yes

revant (changes requested):
LGTM, I added https://hub.docker.com/r/frappe/assets-builder

Docs for install-app usage in custom app README?
That will solve how to install multiple apps from local COPY as well as `git clone`

revant (approved):
merge?

vrslev:
Yep!

Issue #722, opened 2022-03-21
builds failing


revant:
- .env update failed https://github.com/frappe/frappe_docker/runs/5622481846?check_suite_focus=true
- develop build failed, there is no "latest" branch https://github.com/frappe/frappe_docker/runs/5620438636?check_suite_focus=true

@vrslev

revant:
@vrslev develop still failing, sticker is red.

revant:
Why is it trying to pull latest branch on frappe? Frappe doesn't have latest branch, it needs to pull develop branch

```
 > [frappe-socketio builder 2/2] RUN git clone --depth 1 -b latest https://github.com/frappe/frappe /opt/frappe:
#25 0.187 Cloning into '/opt/frappe'...
#25 0.643 warning: Could not find remote branch latest to clone.
#25 0.643 fatal: Remote branch latest not found in upstream origin
------
Dockerfile:4
--------------------
   2 |     
   3 |     ARG FRAPPE_VERSION
   4 | >>> RUN git clone --depth 1 -b ${FRAPPE_VERSION} https://github.com/frappe/frappe /opt/frappe
   5 |     
   6 |     
--------------------
error: failed to solve: process "/bin/sh -c git clone --depth 1 -b ${FRAPPE_VERSION} https://github.com/frappe/frappe /opt/frappe" did not complete successfully: exit code: 128
```

@vrslev where to find the code? FRAPPE_VERSION seems to be set to develop branch by default yet it is trying to pull latest?

develop nightly builds are stuck because of this fail. I tried to debug but if no response I'll fix/hack something up this weekend.

vrslev:
I will try to fix the issues today.

Development build fails because test suite modifies environment variables. Can we merge #705 before fixing the issue?

PR #728, opened 2022-03-24
Fix example.env workflow failures

vrslev:
It failed when Frappe and ERPNext both released new versions. Now it shouldn't fail. To be sure, we need to test it in production.


vrslev:
#722

Issue #128, opened 2020-03-07
Cannot INSTALL_APP


revant:
```
craft@delete-me:~/frappe_docker$ docker exec -it     -e "SITE_NAME=erp.revant.me"     fbench_frappe-python_1 docker-entrypoint.sh new                              
  File "/home/frappe/frappe-bench/commands/new.py", line 9
    install_apps = if os.environ.get("INSTALL_APPS").split(",") else []


```

Images are unusable @chabad360 

even frappe image cannot be installed.

Next time i'm not merging the code. Someone else will

PR #733, opened 2022-03-24
Add platforms to bake file to include m1 (arm64)

restlessronin:
I'm on an M1 mac and I notice that the `bench` image getting pulled is an `amd64` image. I've had some problems with QEMU on my mac in the past, so it would seem better to have a native image.

I'm not really adept at GitHub actions, but it seemed like the fix was straightforward, so I've patched the `docker-bake.hcl` file with what seemed needed to include `arm64`.  I also added the qemu github action.

At the moment, multi architecture works for `build-bench.yml`. Strangely, it isn't working for the `build-stable.yml` and `build-develop.yml`. It's not clear why it should work in the one case, but not the other. I don't understand enough about github actions to be able to diagnose the problem.

However, I think it's important to get this fixed. I've noticed most of the projects I work on have enabled this over the past year, and it's something of a marker of an active, well-supported project.

Not sure how to proceed here. Any suggestions for how to fix, or should this PR be closed?


restlessronin:
spoke too soon. looks like the build process is failing. let me look a little further into this.

restlessronin:
I've made a couple of changes that might fix the broken builds. Build isn't happening, some kind of limit on automatic building of PR's?

restlessronin:
Adding the qemu action allows multi architecture builds for `build-bench.yml`. However this isn't working for the `build-stable.yml` and `build-develop.yml`. It's not clear why it should work in the one case, but not the other.

I don't understand enough about github actions to be able to diagnose the problem.

However, I think it's important to get this fixed. I've noticed most of the projects I work on have enabled this over the past year, and it's something of a marker of an active, well-supported project.

Not sure how to proceed here. Any suggestions for how to fix, or should this PR be closed?

docker-bake.hcl
```diff
@@ -19,11 +19,19 @@ target "bench" {
     context = "images/bench"
     target = "bench"
     tags = ["frappe/bench:latest"]
+    platforms = [
```
vrslev, line 22: Add platforms directive to “default-args” target to prevent repetition

restlessronin: Done. Good Idea. Thanks. It eliminates the repetitions.

As expected, it doesn't change the outcome of the build.

restlessronin:
I have separted the changes to `docker-build-push` into a separate pull request (#737 ). Once that is merged into main, this PR should build correctly.

revant:
v12 build failing? anything to do with py 3.7?

restlessronin:
> v12 build failing? anything to do with py 3.7?

this seems to be the error causing the build failure

#128 425.8 error An unexpected error occurred: "https://registry.yarnpkg.com/rxjs/-/rxjs-6.6.3.tgz: ESOCKETTIMEDOUT".

Looks like a missing yarn package. When I use the browser to download the link manually, I get this JSON

`{"code":"ResourceNotFound","message":"/rxjs/-/rxjs-6.6.3.tgz: does not exist"}`

So it's not just a bad connection during build. The odd thing is that this looks like it has nothing to do with "arm64" (l think that's the reactive JS lib), so I'm not sure why it's the arm64 build that is failing. 

However, when I did a search, I found (an older report)

https://github.com/yarnpkg/yarn/issues/8242

Since manually downloading the link is failing, I don't think the above referenced thread is the correct explanation, but I'm including it anyway.

However, if that is the problem, it could be solved by retrying. The following action was suggested in one of the comments in the thread above.

https://github.com/nick-fields/retry

restlessronin:
I just tried downloading that link again, and it worked this time. Perhaps it was just a temporary npm outage. Is there some way to rerun just the "v12" build?

restlessronin:
The build is still failing apparently due to intermittent outage on yarn registry

https://registry.yarnpkg.com/rxjs/-/rxjs-6.6.3.tgz

is sometimes available and sometimes not (based on manual download in web browser).

Assuming this is a problem that will go away, we don't have to do anything but wait a day or two. If it's an ongoing problem, we have to incorporate some kind of retry code into the build.

vrslev:
v12 build takes 2 hours. It is unacceptable for PR builds. When I introduced Docker Buildx Bake to the project, initially I wanted to build for ARM (I have M1 Mac) but gave up on that because of build time.
Is there a way to reduce build time for ARM?

restlessronin:
> v12 build takes 2 hours. It is unacceptable for PR builds. When I introduced Docker Buildx Bake to the project, initially I wanted to build for ARM (I have M1 Mac) but gave up on that because of build time. Is there a way to reduce build time for ARM?

Not that I'm aware of, but I don't consider myself very well informed about either Docker or Github actions.

For PR builds, is there a way of having the build be manual instead of automated? and more fine grained? so instead of running on every PR commit, it only runs when manually triggered, and one file can be built at a time if necessary?

I think having multi arch images available to be pulled from docker hub is quite important.

So perhaps, it's possible to turn off the automatic builds on PR's? or turn off multi-arch builds for PR? The important thing is that they're available for users of containerized ERPNext.

restlessronin:
> v12 build takes 2 hours. It is unacceptable for PR builds. When I introduced Docker Buildx Bake to the project, initially I wanted to build for ARM (I have M1 Mac) but gave up on that because of build time. Is there a way to reduce build time for ARM?

Not sure if this applies to GitHub - https://github.com/docker/buildx#building-multi-platform-images, but there is an option to build on native nodes.

vrslev:
You can try to enable multi arch build only on "Push" step in action in `docker-build-push.yml`.
https://github.com/docker/bake-action#customizing

We don't use `docker/build-push-action` but here's an example of what I'm talking about: https://github.com/docker/build-push-action/blob/master/docs/advanced/multi-platform.md

revant:
If we can't build arm64 for v12 then we can do it from v13 onwards and skip v12 for arm64.

revant:
Develop build / build / Build (pull_request) Successful in **157m**
Stable build / v12 / Build (pull_request) Successful in **134m**
Stable build / v13 / Build (pull_request) Successful in **169m**

157m, 134m, 169m is too long I think.

@vrslev what do you think?

restlessronin:
The UI for manually triggering CI workflows seems to have been removed.

Therefore, instead of removing the PR trigger, II have commented it out. It can be toggled on and off to "manually" trigger the build.

The remaining triggers, including `workflow_dispatch` and `cron` jobs remain as is.

vrslev:
> Therefore, instead of removing the PR trigger, II have commented it out.

Nothing was removed! We need PR builds

docker-bake.hcl
```diff
@@ -58,10 +62,7 @@ target "default-args" {
         # If `ERPNEXT_VERSION` variable contains "v12" use Python 3.7. Else — 3.9.
         PYTHON_VERSION = can(regex("v12", "${ERPNEXT_VERSION}")) ? "3.7" : "3.9"
     }
-    platforms = [
-        "linux/amd64",
-        "linux/arm64"
-    ]
+    platforms = notequal("pull_request", ${GITHUB_EVENT_NAME}}, ["linux/amd64","linux/arm64"], ["linux/amd64"])
```
vrslev: What if somebody (me) wants to build images locally on `linux/arm64` platform? I don't we should bring GitHub related stuff to `docker-bake.hcl`. Let's not mess up with abstraction layers.

My suggestion is to:
- add `platform` input to `docker-build-push` resuable action
- set `platform` input in v12, v13 builds

Reference: https://github.com/docker/bake-action#inputs

vrslev:
> 157m, 134m, 169m is too long I think.

@revant Yes! I think it is not worth it.
I think for now (until GitHub starts providing ARM machines for GitHub Actions) we abandon the idea of multi arch builds for our main images.
Although we _can_ start providing `frappe/bench` image for `linux/arm64`, it is not that costly.

revant:
> Although we can start providing frappe/bench image for linux/arm64, it is not that costly.

Yes, if it is not costly. That'll solve the issue for Mac based developers. No need to focus arm64 based servers yet.

docker-bake.hcl
restlessronin: i agree this is a cleaner solution, but i'll have to put a little work into figuring it out. give me a day to see what i can come up with.

restlessronin: I wound up using a variation on the above. I added an environment variable `MULTI_ARCH_BUILD` to the `docker-bake.hcl` file, with a default value of `false`.  In the `build_bench.yml` this is set to false for PRs (true otherwise).

I wasn't able to test properly on "stable" and "develop" because it requires replacing the `docker-build-push.yml` on `main`. As it stand now, multi arch can be toggled for `bench` but is disabled for `develop` and `stable`.

To allow toggling on `develop` and `stable`, `docker-build-push` on `main` has to be replaced. I have deliberately not made those changes as yet, they are relatively straightforward, but probably need to be synchronized with simultaneous changes in 'develop' and 'stable' (otherwise the build in `main` may stop working).

restlessronin:
I have added a (not yet tested) commit to allow toggling of multi arch builds on `develop` and `stable`. The builds are multi arch, except on PR.

docker-bake.hcl
restlessronin: have replaced the env var with an input for the `docker-build-push` action because env vars were not being accepted in the job YAML for the action.

restlessronin:
I figured out how to test the `docker-build-push` by linking the action to the branch in my fork. now everything is working as expected. I have linked the action back to the main branch now that the builds have been tested.

@revant , @vrslev ,  I have set `stable` multiarch flag to be false in all situations. `bench` and `develop` are single arch builds for PRs only.

If this is the desired behaviour, the branch is good to go.

docker-bake.hcl
restlessronin: after numerous experiments, I finally figured out how to do what you suggested (by building against an action in a branch in my repo). that's more or less what I wound up implementing.

restlessronin:
@vrslev @revant , long story short, i wound up implementing the featur in the way that was suggested by @vrslev . everything should be working correctly now.

.github/workflows/docker-build-push.yml
```diff
@@ -52,6 +56,7 @@ jobs:
         uses: docker/bake-action@v1.7.0
         with:
           push: true
```
vrslev, line 58: Set platform on Push step. This way images will be build and tested on default platform (`linux/amd64`) and pushed for multiple platforms.

.github/workflows/build_stable.yml
```diff
@@ -50,6 +51,7 @@ jobs:
       repo: erpnext
       version: "13"
       push: ${{ github.repository == 'frappe/frappe_docker' && github.event_name != 'pull_request' }}
+      platforms: linux/amd64
```
vrslev, line 54: Set `linux/amd64,linux/arm64`. Otherwise, what is the point of this PR? 😃

.github/workflows/build_develop.yml
```diff
@@ -28,6 +28,7 @@ jobs:
       repo: erpnext
       version: develop
       push: ${{ github.repository == 'frappe/frappe_docker' && github.event_name != 'pull_request' }}
+      platforms: ${{ github.event_name != 'pull_request' && 'linux/amd64,linux/arm64' || 'linux/amd64' }}
```
vrslev, line 31: This platforms will only affect Push step, so you can remove the condition here and allow both `linux/amd64` and `linux/arm64`.

.github/workflows/build_bench.yml
```diff
@@ -22,13 +22,20 @@ jobs:
...
+          platforms: all
+
       - name: Setup Buildx
         uses: docker/setup-buildx-action@v1
 
       - name: Build and test
         uses: docker/bake-action@v1.7.0
         with:
           targets: bench-test
+          set: "*.platform=${{ github.event_name != 'pull_request' && 'linux/amd64,linux/arm64' || 'linux/amd64' }}"
```
vrslev, line 38: Set platforms on Push step and drop the condition.

revant:
@vrslev merge whenever review is complete, I approve it as long as we've reasonable build durations.

.github/workflows/docker-build-push.yml
restlessronin: done

.github/workflows/build_stable.yml
restlessronin: done

.github/workflows/build_develop.yml
restlessronin: done

.github/workflows/build_bench.yml
restlessronin: done

vrslev:
Last thing. We agreed not to build arm images for 12 version.

restlessronin:
> Last thing. We agreed not to build arm images for 12 version.

done

vrslev:
@revant On PRs build will be fast, on push — not so fast. https://github.com/frappe/frappe_docker/actions/runs/2085442268

PR #742, opened 2022-03-29
Remove git reference from install-app

### LINKED ISSUE #735: install-app from worker image requires git
if install-app with git url is used it doesn't work. It needs git command installed.

should we keep the git clone out of image building and just use COPY to get the code into image?

vrslev:
Before, custom app builders could install apps from Git, this prevents it. Issue was that final backend images (frappe and erpnext) didn't have Git installed. In #735 we decided that we should keep Git out of the backend image.

In this PR:
- Remove Git reference from install-app scripts, update Dockerfiles accordingly
- Update custom app guide
- Fix issue with missing sites/assets folder

Fixes #735.

### DIFF
--- a/images/nginx/install-app.sh
+++ b/images/nginx/install-app.sh
@@ -2,14 +2,9 @@
 set -e
 set -x
 
-APP=$1 BRANCH=$2 GIT_URL=$3
+APP=$1
 
 cd /frappe-bench
-
-if test "$BRANCH" && test "$GIT_URL"; then
-  # Clone in case not copied manually
-  git clone --depth 1 -b "$BRANCH" "$GIT_URL" "apps/$APP"
-fi
 
 # Add all not built assets
 cp -r "apps/$APP/$APP/public" "/out/assets/$APP"
@@ -26,4 +21,4 @@ cp -r sites/assets /out
 
 # Cleanup
 rm -rf "apps/$APP"
-rm -rf sites/assets
+rm -rf sites/assets/*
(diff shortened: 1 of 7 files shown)


vrslev:
@revant Review please

images/nginx/install-app.sh
```diff
@@ -2,15 +2,10 @@
...
 
 cd /frappe-bench
 
-if test "$BRANCH" && test "$GIT_URL"; then
-  # Clone in case not copied manually
-  git clone --depth 1 -b "$BRANCH" "$GIT_URL" "apps/$APP"
-fi
-
 # Add all not built assets
 cp -r "apps/$APP/$APP/public" "/out/assets/$APP"
```
revant, line 10: If app doesn't have `public` directory? can this command not fail with non 0 exit?

revant (changes requested):

images/nginx/install-app.sh
vrslev: https://github.com/frappe/frappe_docker/issues/735#issuecomment-1081641829

revant: > Users shouldn't even try to install apps without frontend code, or I am wrong here?

Some apps may not have frontend code. If you create new app it won't have frontend code. If you do bench get-app, it'll run through the script and build no assets. Yet it goes through the yarn commands. `install-app` should not fail for such apps.

revant (approved):
LGTM,

merge when CI completes.

PR #773, opened 2022-04-11
fix: unable to log into Traefik dashboard

Rp70:
### Please provide enough information so that others can review your pull request:

It was not allow to log into Traefik dashboard

### Explain the **details** for making this change. What existing problem does the pull request solve?

The pipelined sed command `sed 's/\$/\\\$/g'` makes $ escaped but docker compose or Traefik doesn't understand it. I can't login into Traefik dashboard without remove the pipelined sed command.

### DIFF
--- a/docs/single-server-example.md
+++ b/docs/single-server-example.md
@@ -65,7 +65,7 @@ Create a file called `traefik.env` in `~/gitops`
 ```shell
 echo 'TRAEFIK_DOMAIN=traefik.example.com' > ~/gitops/traefik.env
 echo 'EMAIL=admin@example.com' >> ~/gitops/traefik.env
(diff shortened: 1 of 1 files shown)


revant:
Try generating passwords few times.

Initially it was what you have changed. Sometimes it misses out the output from that command. I added sed to fix that.

It is a multi step process here : https://dockerswarm.rocks/traefik/#preparation, may be that solves the problem.

I tried to make it into 1 line command using an env file.

The change you made sometimes doesn't work.
It will show in echo on stdout and in env file. When it goes from env file to compose, it messes things up.

I think when the shell sees `${int}`, e.g. `$1`, `$2`, `$8` etc it acts as argument and tries to substitute that further.

Rp70:
I understand what you say. I tried to generate passwords many times. I even tried to replicate with docker compose v1 and v2. I don't use Docker Swarm so I may not see any problems without sed command. So, please just skip this pull request since it may not suitable to everyone. Thank you for your feedback!

Issue #772, opened 2022-04-11
Improve build times


revant:
```
#134 2520.5   Building wheel for pandas (pyproject.toml): still running...
```

This takes infinite time, with every python version change we face this.

Is there any suggestion, workaround, hack, alternative?

Or we are stuck with it and we can't really do anything about this? Every once in a while pandas is expected to slow down builds? or pandas is expected to slow down builds every time?

revant:
https://github.com/frappe/frappe_docker/runs/5964435466?check_suite_focus=true

this tool 2hrs+

revant:
@vrslev it is building arm64 images!?!

vrslev:
The issue is definitely relevant to arm images. I experience this when build locally on my M1 Mac, too. The only option is to disable builds for arm.

revant:
> The only option is to disable builds for arm?

how? can you guide with GH actions?

i thought only bench image was built for both arm and amd

it seems it published arm images for 13 and 14.

vrslev:
> can you guide with GH actions?

Revert https://github.com/frappe/frappe_docker/pull/733 and https://github.com/frappe/frappe_docker/pull/737 — it should do.

siddhpant:
ARM wheels are available now: https://github.com/pandas-dev/pandas/issues/33971#issuecomment-1253038340

I can just do `pip install pandas` to install it under 10 seconds.

Please build arm64 images, now that the bottleneck does not exist.

PR #737, opened 2022-03-28
add QEMU action

restlessronin:
update build-push separately. once this is merged into main, it can be used by the `images-for-m1-mac` to test the full image build.

Hopefully, it's clear the changes are benign and can be merged.

PR #777, opened 2022-04-11
ci: do not build arm64 images

revant:
Only build arm64 for bench images

for #772

### DIFF
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -14,10 +14,6 @@ on:
       push:
         required: true
         type: boolean
-      platforms:
-        required: true
-        type: string
-        description: "list of target platforms"
     secrets:
       DOCKERHUB_USERNAME:
         required: true
@@ -81,4 +77,3 @@ jobs:
         uses: docker/bake-action@v1.7.0
         with:
           push: true
-          set: "*.platform=${{ inputs.platforms }}"


vrslev (changes requested):
Remove [`platforms` input](https://github.com/frappe/frappe_docker/pull/733/files#diff-b7dce24085c1061be7542e1f68edac2c9002799454a47708546d1e7e6fbf2a10) and ["Setup QEMU" step](https://github.com/frappe/frappe_docker/pull/737/files) from reusable workflow. Everything else is fine.

vrslev (approved):

carstenblt:
Would be great if you would reconsider - it was a great addition and I believe more and more people will use arm instances on AWS, as they are cheaper. I am back to building it locally.

revant:
It won't be considered as long as python builds become faster for pandas and numpy on arm64.
It always takes 1+ hour to build arm64 images.

If you wish to use arm64, you can always build the arm64 images from same code base locally and use it through your custom container registry.

Morakhiyasaiyam:
> Would be great if you would reconsider - it was a great addition and I believe more and more people will use arm instances on AWS, as they are cheaper. I am back to building it locally.

Hey can you guide me how to build it locally!
I tried with  " docker build . "
But throwing dome environment variables requirements but it would be nice if there is script for local build

Morakhiyasaiyam:
Or if you already build arm64 images for yourself then can you send me link for it .now i am using official images of v14.0.2.beta

carstenblt:
I haven't built out myself since the big update. It's interesting, that it takes so long to cross compile. I believe it took like 15 minutes on my arm MacBook.

I will build a new version soon. I can push it to a public docker repo, but I won't maintain it and trusting a stranger on the internet is always a bad idea ;)

Morakhiyasaiyam:
> I haven't built it myself since the big update. It's interesting, that it takes so long to cross compile. I believe it took like 15 minutes on my arm MacBook.
> 
> I will build a new version soon. I can push it to a public docker repo, but I won't maintain it and trusting a stranger on the internet is always a bad idea ;)

I tried to build it on my arm64 device but its crashing device after some commands
@revant is not possible to build arm64 images for not daily but but atleast only after some major update and only for stable and latest v14 ? Once in a while?

revant:
can you try it on custom arm64 vm?

https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners

is this works may be we can think of something.

PR #779, opened 2022-04-12
Add the base URL to the rewrite rules in the nginx config template

polycurve:
When visiting an ERPNext URL with a trailing slash like `https://erp.mydomain.com/app/`, it redirects to an invalid URL `https://erp.mydomain.com:8080/app`.

This happens because the nginx rewrite rule in the configuration template does not specify a base URL for the target. This pull request will fix the invalid redirects.

### DIFF
--- a/images/nginx/nginx-template.conf
+++ b/images/nginx/nginx-template.conf
@@ -54,9 +54,9 @@ server {
 	}
 
 	location / {
- 		rewrite ^(.+)/$ $1 permanent;
-  		rewrite ^(.+)/index\.html$ $1 permanent;
-  		rewrite ^(.+)\.html$ $1 permanent;
+		rewrite ^(.+)/$ https://$http_host$1 permanent;
+		rewrite ^(.+)/index\.html$ https://$http_host$1 permanent;
+		rewrite ^(.+)\.html$ https://$http_host$1 permanent;
 
 		location ~ ^/files/.*.(htm|html|svg|xml) {
 			add_header Content-disposition "attachment";


images/nginx/nginx-template.conf
```diff
@@ -54,9 +54,9 @@ server {
 	}
 
 	location / {
- 		rewrite ^(.+)/$ $1 permanent;
-  		rewrite ^(.+)/index\.html$ $1 permanent;
-  		rewrite ^(.+)\.html$ $1 permanent;
+		rewrite ^(.+)/$ https://$http_host$1 permanent;
+		rewrite ^(.+)/index\.html$ https://$http_host$1 permanent;
+		rewrite ^(.+)\.html$ https://$http_host$1 permanent;
```
revant, line 59: don't make this forced https. it will cause problems for local installs

revant (changes requested):

polycurve:
I've replaced the forced HTTPS with `$proxy_x_forwarded_proto`. If this variable is not set, then it defaults to the value of `$scheme`, so this should work for both HTTP and HTTPS.

revant:
LGTM.

@vrslev what do you think?

vrslev:
Blocked by #781

vrslev:
> what do you think?

If we can access sites by `http://127.0.0.1` when testing locally, it is fine. Can we?

revant:
It seems original nginx.conf also should have this problem then?

https://github.com/frappe/bench/blob/eb7e006bc4c9f946e37c8ca9598dd1ac1265bf98/bench/config/templates/nginx.conf#L79-L81

vrslev:
> It seems original nginx.conf also should have this problem then?

I see where you going from. I meant if we create site named `127.0.0.1` then we should be able to access it by `http://127.0.0.1` url

vrslev:
@revant Can you merge main branch into this one to enable CI?

revant:
all good? @vrslev 

it is showing 3 files changed due to the merge. Okay?

revant:
@polycurve any update on native-bench test and 127.0.0.1 test?

polycurve:
> > what do you think?
> 
> If we can access sites by `http://127.0.0.1` when testing locally, it is fine. Can we?

Yes, local testing works fine.

image

PR #793, opened 2022-05-01
Fix builds failing

vrslev:
Fixes #792

PR #794, opened 2022-05-01
ci: bypass docker compose stop failure

revant:
docker compose stop without starting the project
results in failure. Bypass the exception for
docker compose stop


revant:
follow #793

PR #804, opened 2022-05-11
Fix volumes permissions

timemaster5:
Here is a fix for the volumes permission issue I had on Ubuntu 21.10

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -24,6 +24,7 @@ services:
 
   backend:
     <<: *backend_defaults
+    user: "1000:1000"
     volumes:
       - sites:/home/frappe/frappe-bench/sites
       - assets:/home/frappe/frappe-bench/sites/assets:ro
@@ -47,23 +48,28 @@ services:
   websocket:
     <<: *depends_on_configurator
     image: frappe/frappe-socketio:${FRAPPE_VERSION}
+    user: "1000:1000"
     volumes:
       - sites:/home/frappe/frappe-bench/sites
 
   queue-short:
     <<: *backend_defaults
+    user: "1000:1000"
     command: bench worker --queue short
 
   queue-default:
     <<: *backend_defaults
+    user: "1000:1000"
     command: bench worker --queue default
 
   queue-long:
     <<: *backend_defaults
+    user: "1000:1000"
     command: bench worker --queue long
 
   scheduler:
     <<: *backend_defaults
+    user: "1000:1000"
     command: bench schedule
 
 # ERPNext requires local assets access (Frappe does not)


revant:
@vrslev adding the `user: 1000:1000` okay? 

we have `USER frappe` in docker images right?

vrslev:
Do you have any reference of the problem? Cause I don’t think we should make this kind of change “just because”. Also this could break someone’s setup.

revant:
@timemaster5 you can make it part of documentation then?

vrslev:
@revant I didn’t find any explanation from @timemaster5 why exactly setup stopped working on Ubuntu 21.

revant:
I sometimes face this. I don't know why it happens but sometimes volume are only allowed to be accessed by root. In that case I run chown -R 1000:1000 /path/in/vol as root user once. Then the containers get permission to access/write to volume. It rarely happens. It's not everytime.

My guess is, it may be related to docker or kernel being upgraded and machine has not yet restarted the services. I have no clue as it rarely happens and can't be deliberately reproduced.

timemaster5:
> @revant I didn’t find any explanation from @timemaster5 why exactly the setup stopped working on Ubuntu 21.

It didn't stop working. It never worked :) It was a default clean installation.

I know there was some back and forth with permissions of volumes in docker, but I believe this is the correct implementation because it is how it generally works in Linux. 

If part of your setup runs as root, that's fine. But when a custom user (frape in /home/frappe) needs to use some directory previously created as root, the correct permissions of that directory must be set as @revant suggested. 

Suppose you don't need to fix it every time. Then, that should be considered a bug, i.e. in the older Docker versions or incorrectly set permissions of the docker volumes directory, rather than trying to avoid it.

From the basic understanding of a problem, if you create a directory as a root, and then you want to use it as a user, you have to set a correct mod or owner. Instead of relying on some other maintainer of distro/docker/.. to do it just right, and hope that it eventually works.

vrslev:
Assuming that what you’re saying is objectively correct, we should fix the bug in Dockerfiles instead of relying on potentially changeable compose files. Is there a way to do this?

revant:
I rarely face this, whenever I do, I either run an additional container with root priv and fix the volume permissions. OR I directly fix the volume permissions on NFS server or docker host.

revant:
@timemaster5 can we add one more compose override and document it in this case?

`compose.fix-vol-perm.yml` it will override `user` for the necessary container/service.

We document it as a note in case of permission error?

That way nothing changes in main compose and current working setups of users and for those who need the fix can override.

vrslev:
@revant There’re already too many overrides

revant:
I don't think there is a solution for this unless we figure out what exactly happens when there is permission error.

@vrslev close this or merge?

revant:
@skywalker282 

you need to build your custom image from your own Containerfile. Check this comment: https://github.com/frappe/helm/issues/181#issuecomment-1698725817

in breif,

- mounting `sites` will cause `sites/assets` to vanish
- you need to manually re-create `sites/assets` and/or copy things from image to `assets/sites` in your bind-mount on container entrypoint.
- That is similar to using `cri-o` which will create empty `sites` without `sites/assets`, check above link for changes to image.

Issue #836, opened 2022-07-01
use py 3.10 for v14


revant:
https://github.com/frappe/frappe/pull/17355

PR #837, opened 2022-07-03
add python 3.10

revant:
for #836


vrslev:
Bench image build needs to be fixed. Everything else is fine and working.

vrslev (approved):

1ubuntuuser:
Nice. trying out v14 now... ran into `ERROR: Package 'frappe' requires a different Python: 3.9.9 not in '>=3.10'` on bench init but realised I just needed to run `docker pull frappe/bench`

PR #857, opened 2022-07-21
Fix for multiple simultaneous running stacks

Pazdo:

With old version, there was not option to run multiple stacks simultaneously

Changed mariadb-network to bench-network in most services. As result, frontend service doesn't see multiple backend services.

### DIFF
--- a/docs/compose/compose.multi-bench.yaml
+++ b/docs/compose/compose.multi-bench.yaml
@@ -13,6 +13,7 @@ services:
   configurator:
     networks:
       - bench-network
+      - mariadb-network
   backend:
     networks:
       - mariadb-network
@@ -20,21 +21,26 @@ services:
   websocket:
     networks:
       - bench-network
+      - mariadb-network
   scheduler:
     networks:
       - bench-network
+      - mariadb-network
   queue-default:
     networks:
       - bench-network
+      - mariadb-network
   queue-short:
     networks:
       - bench-network
   queue-long:
     networks:
       - bench-network
+      - mariadb-network
   redis:
     networks:
       - bench-network
+      - mariadb-network
 
 networks:
   traefik-public:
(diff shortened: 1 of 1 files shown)


revant:
This file defines the mariadb-network

https://github.com/frappe/frappe_docker/blob/1448e46a9b042b8b32e75913348355ac6d563209/docs/compose/compose.mariadb-shared.yaml

Do the containers are able to connect to db?

Pazdo:
Only `backend` container needs to be connected to db, I don't think any other services needs to be connected to mariadb-network.  I tested it, and everything works.

revant:
workers need to be connected with redis cache, redis queue and MariaDB, try background jobs like sending email, triggering webhooks from frappe framework.

scheduler needs redis-queue and socketio needs redis-queue.
For testing socketio try opening same doc from 2 browsers, socketio should notify someone else is viewing the document.

for scheduler, check if one of the hourly, daily or cron python functions from hooks.py work

Pazdo:
If frontend-worker doesn't need MariaDB, then this config should work. Tested everything you posted, everything works. If there is something more I can test feel free to comment and I will test it. If it wont work, then the easiest solution will be to rename services for every stack based on .env variable.

revant:
I don't use the single server setup using docker. If I use docker, I use this https://gist.github.com/revant/6a67f775aa8198f9b306f32d6872799f
So I won't ever use what is mentioned here. I can merge this by trusting you. 

But if something fails for someone else, I have to answer the questions, no one else help me in there. Can you add any tests, at least that will be determisintic and help me filter out real issues vs beginners issues.

revant:
can anyone else test this PR? It'll take some time for me to test it.

revant:
I trust your effort @Pazdo

Anyway if there are questions no one else is going to answer.

Issue #896, opened 2022-08-24
Remove v12 builds


revant:
Based on https://discuss.erpnext.com/t/upgrade-to-version-13-or-newer-to-continue-receiving-regular-fixes/93052

- Remove version 12 production builds

Should python 3.7 be removed from frappe/bench:latest image?

revant:
- [ ] remove v12 production builds

revant:
#922

PR #898, opened 2022-08-25
feat: remove version-12 dependencies

FHenry:
As https://discuss.erpnext.com/t/upgrade-to-version-13-or-newer-to-continue-receiving-regular-fixes/93052
version-12 is no more supported, we can remove it from docker bench image

closes : #896

### DIFF
--- a/images/bench/Dockerfile
+++ b/images/bench/Dockerfile
@@ -44,7 +44,6 @@ RUN apt-get update \
     libwebp-dev \
     redis-tools \
     rlwrap \
-    libbz2-dev \
     tk8.6-dev \
     ssh-client \
     # VSCode container requirements
--- a/images/bench/Dockerfile
+++ b/images/bench/Dockerfile
@@ -13,9 +13,6 @@ RUN apt-get update \
     postgresql-client \
     gettext-base \
     wget \
-    # for ERPNext v12
-    # TODO: Remove after v12 is deprecated
-    python2 \
     # for PDF
     libssl-dev \
     fonts-cantarell \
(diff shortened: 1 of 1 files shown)


revant:
should we remove it from bench image?
people may be using it to develop their v12 custom apps?

FHenry:
Hi @revant 
Yes I wonder also...
I made this PR because I see the announcement on the forum, and as I play with bench imagea lot, I think it can be good to have only "supported" inside.

May be should I update the PR with another dockerfile  and add some documentation for older version ?

What do you think ?

revant:
> May be should I update the PR with another dockerfile and add some documentation for older version ?

Should we just tag and push existing bench as `version-12` tag, then deprecate code and continue with daily `latest` builds. People who want to develop version-12 will stick to `version-12` tag.

FHenry:
Sounds very good.
Tag version-12 is actual one and merge this PR for lastest ?
Only you car create tag can do that (and I'm happy with that ;-))

revant:
Tag is good, We may also need branches like `version-10`. I'll create those. Docker hub has some policy where they remove image tags that are not used (not sure about latest updates). So if we tag there is a chance of the image tag to be deleted. In that case branch will help users build their own images.

revant:
I've created `version-12` branch. Manually building and pushing `frappe/bench:version-12`

FHenry:
Nice, so we can go for `main` branch only version-13 and version-14 by merging this pr ?

revant:
image: frappe/bench:version-12 is available now.

PR #899, opened 2022-08-25
fix: same max_allowed_packet for MariaDB engine as for mysqldump utilities

FHenry:
On some old ERPNext instance tabDeleted Documents have a large amount of data into data column and `bench backup` that use mysqldump cannot perform a correct dump.

The max-allowed-packed on "services : mariadb" in docker-compose only apply on engine instance not on mysqldump.

It's should be better to have the same limit on both side


revant:
MariaDB is not part of bench image. Is /etc/mysql available in bench image?

Tests needs to be green.

FHenry:
Ok, my bad, I should have test build before PR.... Too confident....

I keep that in my mind and change it on my local image, We should make a docker file for mariadb for this, the use-case are really few

PR #922, opened 2022-09-10
Deprecate assets builder

revant:


revant:
`Error: no such option: --production` for frappe framework v13. Option exists for version 14

PR #944, opened 2022-09-29
fix: set traefik Host in backticks

RafaelKr:

The `Host` must be inside backticks: https://doc.traefik.io/traefik/routing/routers/#rule

This prevents `Error while parsing rule Host(sub.sub.example.com), error: *ast.BinaryExpr is not supported`


revant:
what if

```
SITES=`site1.com`,`site2.com`,`site3.com`
```

with this change will it become?

```
SITES=``site1.com`,`site2.com`,`site3.com``
```

RafaelKr:
You're right, I didn't read [the docs](https://github.com/frappe/frappe_docker/blob/main/docs/single-server-example.md) correctly and missed the backticks which should be in the env file.

revant:
I think if backticks don't work you can use escaped double quotes (`\"`)

RafaelKr:
They work, I just didn't see that they should be added in the env file :)

PR #961, opened 2022-10-15
fix: install dependencies for v13 on creation

vorasmit:

Added dependencies necessary to install version 13 of Frappe & ERPNext (For Dev Container)

The objective is to spin off the container without worrying about installing basic dependencies.

### DIFF
--- a/devcontainer-example/devcontainer.json
+++ b/devcontainer-example/devcontainer.json
@@ -14,6 +14,7 @@
   "dockerComposeFile": "./docker-compose.yml",
   "service": "frappe",
   "workspaceFolder": "/workspace/development",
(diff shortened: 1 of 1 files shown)


revant:
recently i faced this.
when we try to create bench with non global python it needs virtualenv.

we've python through pyenv and not through os repos. that way we can control exact version of python needed for development. getting it from debian sources will not be compatible with benches.

what worked for me in case of v13 bench was:

```shell
pyenv global 3.9.9
pip install virtualenv
bench init ... # no need to specify python as we've set 3.9.9 to global
```

after you're done with creating bench you can set global back to 3.10

vorasmit:
I understood what you meant.
However, I am new to docker and am not confident about how it's set up here. So I will leave it to you. Thanks 😄.

PR #975, opened 2022-10-20
feat: make proxy read timeout configurable via env

WebsiteDeveloper:
This solves issues with multi document Print, which might take more than 2mins and allows the behaviour to be overridden while keeping the sane defaults

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -35,6 +35,7 @@ services:
       UPSTREAM_REAL_IP_ADDRESS: ${UPSTREAM_REAL_IP_ADDRESS:-127.0.0.1}
       UPSTREAM_REAL_IP_HEADER: ${UPSTREAM_REAL_IP_HEADER:-X-Forwarded-For}
       UPSTREAM_REAL_IP_RECURSIVE: ${UPSTREAM_REAL_IP_RECURSIVE:-off}
+      PROXY_READ_TIMOUT: ${PROXY_READ_TIMOUT:-120}
     volumes:
       - sites:/usr/share/nginx/html/sites
       - assets:/usr/share/nginx/html/assets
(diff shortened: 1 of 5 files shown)


revant:
This will also need changes in helm chart or they will start failing

- https://github.com/frappe/helm/blob/2ef742cca761d68e3b43d5691f6ef810bcf1d7d2/erpnext/templates/deployment-nginx.yaml#L28
- https://github.com/frappe/helm/blob/2ef742cca761d68e3b43d5691f6ef810bcf1d7d2/erpnext/values.yaml#L21

WebsiteDeveloper:
I changed the PWD and ran the pre-commit.
About the helm chart: i'm not really sure how to add that there.

Concerning breaking current deployments the only way i would currently see is modifying the entrypoint to set the value as a default before envsubset replaces the config env variables.
I will take a look at that next week.

revant:
> Concerning breaking current deployments the only way i would currently see is modifying the entrypoint to set the value as a default before envsubset replaces the config env variables.

If you manage to get this working then set all the defaults there. It won't be breaking change then. There is a right way to do it as per the https://github.com/nginxinc/docker-nginx-unprivileged env / entrypoint sequence.

revant:
https://discuss.erpnext.com/t/how-to-upload-large-file-size-more-than-200mb/98344
this also needs to configure `client_max_body_size`

compose.yaml
```diff
@@ -35,6 +35,7 @@ services:
       UPSTREAM_REAL_IP_ADDRESS: ${UPSTREAM_REAL_IP_ADDRESS:-127.0.0.1}
       UPSTREAM_REAL_IP_HEADER: ${UPSTREAM_REAL_IP_HEADER:-X-Forwarded-For}
       UPSTREAM_REAL_IP_RECURSIVE: ${UPSTREAM_REAL_IP_RECURSIVE:-off}
+      PROXY_READ_TIMOUT: ${PROXY_READ_TIMOUT:-120}
```
ankush, line 38: Does increasing this alone really help when gunicorn timeout is 120 seconds?

https://github.com/frappe/frappe_docker/blob/45c6401da9083b4b53dc7660c5303f339e77e73e/images/worker/Dockerfile#L123

revant: https://discuss.erpnext.com/t/erpnext-server-keeps-timing-out/90555/7

gunicorn can be overriden at runtime

WebsiteDeveloper:
@revant I updated the pr with an ENV statement and tested it locally
Seems to work like this when the env is not set.
Can you verify?

PR #992, opened 2022-11-02
feat: install client apps using script

vorasmit:

- Created install scripts for developers.
- Client-centric, where the script can create a new bench/site for the client.
- Configure apps for the client in apps.json, similar to the example file.
- Added relevant documentation.

edit: 
pre-commit lint issues fixed. also marked installer.sh as executable.
shellcheck `SC1091` won't fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@ sites
 
 development/*
 !development/README.md
+!development/installer.sh
+!development/apps-example.json
 !development/vscode-example/
 
 # Pycharm
--- a/devcontainer-example/docker-compose.yml
+++ b/devcontainer-example/docker-compose.yml
@@ -36,6 +36,8 @@ services:
       - SHELL=/bin/bash
     volumes:
       - ..:/workspace:cached
+      # Enable if you require git cloning
+      # - ${HOME}/.ssh:/home/frappe/.ssh
     working_dir: /workspace/development
     ports:
       - 8000-8005:8000-8005
--- a/development/README.md
+++ b/development/README.md
@@ -61,6 +61,8 @@ Notes:
 
 ### Setup first bench
 
+> Jump to [scripts](#setup-bench--new-site-using-script) section to setup a bench automatically. Alternatively, you can setup a bench manually using below guide.
+
 Run the following commands in the terminal inside the container. You might need to create a new terminal in VSCode.
 
 ```shell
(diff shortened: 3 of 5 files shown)


athul:
Hi, I just tried to do it with the installer script but failing saying `WARN: Command not being executed in bench directory`

athul:
Honestly, My 2 cents would be to replace the bash script with Python. 

Python would be more saner than bash.

 And rather than having a client name we could just specify the apps in the json file. Kinda like this

```json
{
[
"name":"frappe",
"branch":"develop",
...
...
],
[
…
]
}
```
Like this. As far I could see, the "client_name" key isn't used anywhere.

I'm currently writing the Python script for the same,

revant:
`apps.json` taken as input to `bench init --frappe-branch=version-14 --apps_path=apps.json --skip-redis-config-generation frappe-bench` looks like this:

```json
[
  {
    "url": "https://github.com/abc/xyz",
    "branch": "main"
  },
  {
    "url": "ssh://git@gitlab.com/abc/pqr.git",
    "branch": "main"
  },
  {
    "url": "ssh://git@gitlab.com/abc/bank.git",
    "branch": "master"
  },
  {
    "url": "ssh://git@gitlab.com/abc/admin.git",
    "branch": "main"
  },
  {
    "url": "ssh://git@gitlab.com/abc/cash.git",
    "branch": "main"
  }
]
```

vorasmit:
apps.json takes client name too.

This is introduced, as most clients use different sets or apps on different version.

So, you can maintain here configuration client-wise.

athul:
What is client here?

vorasmit:
Client => Customer

revant:
frappe-bench just supports array of objects e.g. `[{"url":"http_or_ssh_url","branch":"version-14"}]`

code:

https://github.com/frappe/bench/blob/34197056ea7de6735ea0fd8fc63b18093b3f634b/bench/app.py#L688-L696

vorasmit:
I didn't something like this existed in frappe-bench.
Are you using the same JSON for production and development?

Do you suggest we use `url` instead of `upstream` for consistency?

revant:
I'm using this internal bench code to initiate a bench in new production images

check this doc https://github.com/revant/frappe_docker/blob/refactor-single-image/docs/custom-apps.md

PR #1004, opened 2022-11-28
feat: Add Docker compose files with Version and Sitename as variables

athul:
Adds two compose files
- erpnext.yml for ERPNext setup
- frappe.yml for FrappeFramework instance
- Also adds an .env.example for the environment variable names

This will be used in the new easy install script :)


revant:
Why duplicate all contents of yml? In case of ERPNext only few images or services can be overridden.

compose-files/.env.example
```diff
@@ -0,0 +1,2 @@
+VERSION=v14
+SITENAME="site1.local"
```
revant, line 2: remove `.env.example` file add / use / change variables from

https://github.com/frappe/frappe_docker/blob/main/example.env

You can add SITENAME in example.env `FRAPPE_VERSION` and `ERPNEXT_VERSION` already exists there. You can use one from it instead of adding 1 more `VERSION`. Also example.env is updated on every release and the *_VERSION(s) are bumped as per the upstream release. You don't need to maintain that env further.

compose-files/README.md
```diff
@@ -0,0 +1,5 @@
+These compose files are forks of the the [pwd.yml](/pwd.yml) file with some changes. The most notable changes being selecting the version to be installed and the sitename to be created. These are provided with environment variables. Please refer [.env.example](/compose-files/.env.example) for the environment variable naming.
```
revant, line 1: docs related to .env.example will change if file is removed.

revant (changes requested):

athul:
Closing this as it's not a requirement

PR #1044, opened 2023-01-17
feat: add restic

revant:
allows incremental snapshot backups
remove custom push-backup script


revant:
@sdfateh any comments?

revant:
Removing .git from frappe reduces 17MB, removing .git from erpnext removes 19MB and from payments it removes 232K.

Adding restic adds 24MB.

Merging.

PR #1053, opened 2023-01-20
add read perms to nginx-entrypoint

sdfateh:


images/custom/Containerfile
```diff
@@ -64,7 +64,7 @@ RUN useradd -ms /bin/bash frappe \
     && chown -R frappe:frappe /var/log/nginx \
     && chown -R frappe:frappe /var/lib/nginx \
     && chown -R frappe:frappe /run/nginx.pid \
-    && chmod +x /usr/local/bin/nginx-entrypoint.sh
+    && chmod 755 /usr/local/bin/nginx-entrypoint.sh
```
put3r-r00t3r, line 68: Please also add "&& chmod 644 /templates/nginx/frappe.conf.template" after line 67

images/custom/Containerfile
```diff
@@ -64,7 +64,7 @@ RUN useradd -ms /bin/bash frappe \
...
```
put3r-r00t3r, line 67: Add slash at the end to allow for next line "&& chmod 755 /usr/local/bin/nginx-entrypoint.sh \"

put3r-r00t3r (changes requested):
@sdfateh 

Based on the main frappe/erpnext image, here are what the permissions and ownership that "should" be on those two files after a successful build:

```
$ docker run --rm -it frappe/erpnext:v13.43.2 /bin/bash
...docker pull output omitted for brevity...
frappe@fa52099bb6e2:~/frappe-bench$ ls -l /usr/local/bin/nginx-entrypoint.sh
-rwxr-xr-x 1 root root 1502 Jan 18 06:01 /usr/local/bin/nginx-entrypoint.sh
frappe@fa52099bb6e2:~/frappe-bench$ ls -l /templates/nginx/frappe.conf.template
-rw-r--r-- 1 root root 3239 Jan 18 06:01 /templates/nginx/frappe.conf.template
```
So I think these two lines (as I commented on lines 67 & 68) should be all that are needed:
```
&& chmod 755 /usr/local/bin/nginx-entrypoint.sh \
&& chmod 644 /templates/nginx/frappe.conf.template
```
This works for me on my system and nginx no longer crashes, which allows the frontend service container to stay running and serve the site.

images/custom/Containerfile
sdfateh: Why 644 for this file?
What problem are you facing?

revant: I think there is no problem as long as the template available for user frappe to read and generate config from.

The comment on the PR states the permissions that are in frappe/erpnext image and the `chmod 644` will make sure the permission is set to same.
 
- Original image doesn't do this acrobatics. It is built with linux (gh actions ubuntu runners). Use linux runners or linux machines to avoid additional workarounds. May be there are hidden ones yet to come?

revant:
I didn't check if it works. It should work. @put3r-r00t3r inform if it does.

put3r-r00t3r:
Yes, that's correct @revant. After I fixed the entrypoint I got an almost identical error where nginx crashes and the frontend container stops. The docker frontend container logs reported the permissions error reading the template file. The permissions were set to 600 which means the frappe user could not read it:

```
$ docker compose -p frappe_docker -f docker-compose.yml logs frontend
frappe_docker-frontend-1  | PROXY_READ_TIMEOUT defaulting to 120
frappe_docker-frontend-1  | /usr/local/bin/nginx-entrypoint.sh: line 49: /templates/nginx/frappe.conf.template: Permission denied
```

```
$ docker compose -p frappe_docker -f docker-compose.yml run --rm -it frontend ls -l /templates/nginx/frappe.conf.template
...
-rw------- 1 root root 3239 Jan 19 20:27 /templates/nginx/frappe.conf.template

```

I will pull the changes and let you know if they work.

put3r-r00t3r:
Confirmed this is working. I agree, @revant this should not be necessary, but it seems to be a side effect of "git pull" and the Containerfile's "COPY" commands:

```
$ grep COPY images/custom/Containerfile
COPY resources/nginx-template.conf /templates/nginx/frappe.conf.template
COPY resources/nginx-entrypoint.sh /usr/local/bin/nginx-entrypoint.sh
COPY --from=builder --chown=frappe:frappe /home/frappe/frappe-bench /home/frappe/frappe-bench
```

```
$ ls -l resources/
total 8
-rwx------ 1 tccadmin users 1502 Jan 19 12:27 nginx-entrypoint.sh
-rw------- 1 tccadmin users 3239 Jan 19 12:27 nginx-template.conf
```

Those container permissions in the images that are built appear to be such because that's how they exist on my file system after I clone or pull the frappe_docker repository from github.

revant:
How does the main image is built and works? Same files, same Containerfile, same repo.

At least new copy-pasters using windows can blindly copy paste without complaining (or understanding) permissions.

Issue #876, opened 2022-08-12
version 14 - Install officials custom apps


guilmori:
It has been a while since I’ve been here, I see the documentation has greatly improved, good job 👍 

Now that version 14 is out, and many modules have been extracted to custom apps, I’d like to see how to install a couple of custom apps, following this [great tutorial ](https://github.com/frappe/frappe_docker/blob/main/docs/single-server-example.md) 

I saw this [custom app doc](https://github.com/frappe/frappe_docker/tree/main/custom_app), but it seems targeting how to build our own custom app images.

revant:
Right now there is no official image with all apps.
You need to build your own image with required official as well as custom apps.

Using the official docs, I have example repo here https://github.com/castlecraft/custom_frappe_docker

Any suggestions you have to make it even easier?

guilmori:
Ah I see. 

As per installation instructions of most/all custom apps, I thought the usual get-app / install-app was to be executed somehow on startup of container, but this isn’t probably the docker way.

sdfateh:
@revant @guilmori 

I suggest that we make an image for each app, and an image that collects all the applications, in this way, whoever needs one application can get it, and whoever needs more than one application takes the general image and install the required application directly on the site.

guilmori:
Your comment really puzzle me… how can Frappe recommend docker as de facto method of installation when they don’t maintain proper images ?

revant:
They recommend frappecloud or some plan that'll support your self hosted environment. I've not checked the latest offer. And if you go for self-host plan (if it exists) most probably it'll be manual setup on VM.

I prefer containers with all its best practices, so I shared my setup with community. Container using community helped in improving it and getting this repo where it is. My update fails, I fix and contribute. Others do the same. Rest there are open issues that no one has answers yet.

Most of the veterans don't use containers. For newcomers who have already used containers, this repo will make sense. Newcomers who have not used containers but get lost in 50 step rituals turn to 4-5 step container setup rituals and may have more probability of success. 

Recommending docker for ritual following newcomer is better than recommending manual setup.

Why you'll get free support from me? 1) I use it. 2) Customers of [company that I own](https://www.castlecraft.in) use it.

revant:
`frontend.Dockerfile` that works with `version-14`

```dockerfile
# syntax=docker/dockerfile:1.3

ARG FRAPPE_VERSION
ARG ERPNEXT_VERSION

FROM frappe/bench:latest as assets

[...]

PR #1062, opened 2023-01-27
Clarification on how to use custom image

maovidal:
This may help a bit to clarify what additional changes are expected when using custom images.
This was taken after the post https://github.com/frappe/frappe_docker/issues/876#issuecomment-1261834758

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -1,3 +1,9 @@
+x-customizable-image: &customizable_image
+  # By default the image used only contains the `frappe` and `erpnext` apps.
+  # See https://github.com/frappe/frappe_docker/blob/main/docs/custom-apps.md
+  # for using custom images
+  image: frappe/erpnext:${ERPNEXT_VERSION:?No ERPNext version set}
+ 
 x-depends-on-configurator: &depends_on_configurator
   depends_on:
     configurator:
@@ -5,7 +11,7 @@ x-depends-on-configurator: &depends_on_configurator
 
 x-backend-defaults: &backend_defaults
   <<: *depends_on_configurator
-  image: frappe/erpnext:${ERPNEXT_VERSION:?No ERPNext version set}
+  <<: *customizable_image
   volumes:
     - sites:/home/frappe/frappe-bench/sites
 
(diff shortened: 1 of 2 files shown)


sdfateh:
You Just add more lines to the files with the same effect.
Why do you think we needed more sections?

revant:
If you change the main compose.yaml also make sure that the new `easy-install.py` script from frappe/bench works.

I'd think of something that automates `easy-install.py` either using `build:` or replacing `image:`

revant:
Atleast fix the lint. I cannot co-commit on your main branch.

maovidal:
> You Just add more lines to the files with the same effect.
> Why do you think we needed more sections?

Besides identifying that the custom images are only for the workers, scheduler, frontend, backend and not websocket/socketio, I think the proposed modifications follow the use of [extension fragments](https://docs.docker.com/compose/compose-file/#using-extensions-as-fragments), which is already used partially on the current compose files. Not only those may be easier to read, but also may help to maintain and extend that kind of components.

maovidal:
> Atleast fix the lint. I cannot co-commit on your main branch.

I'm sorry about that. I did not checked the CONTRIBUTING guide prior my pull request and I haven't had any experience with any of those tools before. But I have already fixed them and performed the tests locally.

What should I do now? Should I commit those changes?

revant:
yes commit the additional changes with commit message `chore: fix lint`

compose.yaml
```diff
@@ -56,7 +62,7 @@ services:
 
   websocket:
     <<: *depends_on_configurator
-    image: frappe/erpnext:${ERPNEXT_VERSION}
+    image: frappe/erpnext:${ERPNEXT_VERSION:?No ERPNext version set}
```
revant, line 65: websocket image can also be custom. it'll have Frappe and socketio

revant:
if you pull the latest changes it'll fix the v14 tests

compose.yaml
revant: @maovidal can you also change this image? No need to keep it. Custom image will have socketio in it.

maovidal: Done.
Hey @revant, maybe not the place nor the time, but certainly I want to express my gratitude to you for all you have shared on this repo. I've learned a lot.

Issue #1060, opened 2023-01-27
Do not have the ERPNext components after docker deployment


joao-guerra:
Hi,

I am trying out ERPNext and followed the instructions and lifted up a docker.
When I open the site I only get a barebones Frappe installation with none of the ERP options available (attached screenshot).
That also happens when I try Play with Docker and I get the same thing as my local installation.

I saw videos that show the ERP options (even the top left logo is an E and I only get an F logo) but I am unable to get them.

What am I missing here?

![Screenshot 2023-01-27 at 12 34 54](https://user-images.githubusercontent.com/2833673/215088435-53e15ed5-36d8-4497-8fa9-11b3b56509a1.png)

revant:
wait for 5-10 min before starting setup wizard

follow logs for create-site service. it should complete.
as soon as frappe is installed site becomes visible. bottom right footer says `frappe`
after ERPNext gets installed footer changes to `ERPNext`

joao-guerra:
I also tried PWD again, I waited 30 minutes and same result.

blackpawninja:
I experience the same issue. `create-site` hangs after `Could not find app "payments"` and ERPNext is not installed even after 3 hours waiting

revant:
Your issue uncovered that payments is now an optional dependency. Removed it from new-site command.
It should work with latest pwd.yml now.

PR #1063, opened 2023-01-28
fix: remove payments

revant:
payments is optional dep for erpnext

https://github.com/frappe/erpnext/pull/33245


revant:
fixes #1060

PR #1076, opened 2023-02-07
Retrieving Assets volume 

sdfateh:
Since the assets directory is inside the `sites` volume, it will keep the first data of initialization.
So If any new files come up with a new image it will not appears and will show an error like this:

image

So we have to retrieve the volume of the asset and do the population when updating.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -55,8 +55,6 @@ services:
       UPSTREAM_REAL_IP_RECURSIVE: ${UPSTREAM_REAL_IP_RECURSIVE:-off}
       PROXY_READ_TIMOUT: ${PROXY_READ_TIMOUT:-120}
       CLIENT_MAX_BODY_SIZE: ${CLIENT_MAX_BODY_SIZE:-50m}
-    volumes:
-      - sites:/home/frappe/frappe-bench/sites
     depends_on:
       - backend
       - websocket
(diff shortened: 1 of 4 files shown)


revant:
i use helm charts and swarm stacks, I'll not be populating assets like this anyway.

if you remove the container it'll remove the temporary assets volume it creates, on new container creation new vol with fresh assets gets created. this is what happens on swarm and kubernetes automatically. names for containers are randomised there. on docker compose container names remain same even after restart, may be causes same temp volumes to get attached again.

if it works for you I'll merge. should anyone else test?

will it affect easy-install? because I don't use that either! and that seems to be only way ERPNext gets installed for self hosted new comers. if your changes affect easy install, issues will land up here.

i land up answering all issues, hope i can tag you on any issues related to easy-install and compose based setup.

revant:
> Since the assets directory is inside the sites volume, it will keep the first data of initialization.

as long as there is `VOLUMES []` mentioned in docker file, nested mounting will be done automatically. we're mounting logs, sites and assets automatically like that.

revant:
@athul can you verify easy-install?

sdfateh:
> as long as there is VOLUMES [] mentioned in docker file, nested mounting will be done automatically. we're mounting logs, sites and assets automatically like that.

This helps a lot since I sometimes don't use `down` for prod for less downtim.
I'll close the PR

sdfateh:
@revant Do you use PWD for a swarm or there is any other full stack definition?

revant:
@sdfateh check this for comprehensive swarm setup

https://github.com/castlecraft/custom_containers/blob/main/docs/docker-swarm.md

blaggacao:
Suggestion: attach volumes per site.

PR #1091, opened 2023-03-06
Update nginx-template.conf

Sajith-K-creater:
Update to allow access of private files

* Changed $http_host to ${FRAPPE_SITE_NAME_HEADER} for ^/protected/


athul (approved):
LGTM. Had faced the same issue. Had to use Docker nginx template to fix the problem

PR #1098, opened 2023-03-13
Build arm64 images in addition to amd64 images

ChillarAnand:


revant:
how much time it takes?

we disabled it because last time it took around 2hrs.

ChillarAnand:
It was taking too long as pandas didn't have an arm64 build earlier. It should work fine now.

On my machine, it takes ~10 minutes.

revant:
> It was taking too long as pandas didn't have an arm64 build earlier.

Hope it remains like that in future after new python and pandas is released.

revant:
Some weird behaviour during push step the image is still being built!

https://github.com/frappe/frappe_docker/actions/runs/4401435320/jobs/7707642818

ChillarAnand:
Let me check if any step is missed CI config.

gavindsouza:
> > It was taking too long as pandas didn't have an arm64 build earlier.
> 
> Hope it remains like that in future after new python and pandas is released.

Side note: Pandas isn't a core dependency on Frappe/ERPNext from v14. Should be fine going forward.

ChillarAnand:
There is no problem with the docker build itself. I am able to build and push arm64 images in ~10 minutes.
https://hub.docker.com/repository/docker/chillaranand/erpnext/tags?page=1&ordering=last_updated

This change is reverted as GitHub action runners are amd64 and the cross-platform build is taking > 1 hour which is causing the build to fail. Need to find a better solution.

ChillarAnand:
As seen from the logs, it is related to yarn install too. Here is a relevant discussion. https://github.com/nodejs/docker-node/issues/1335

PR #1146, opened 2023-06-02
Fix typo in PROXY_READ_TIMEOUT in .env and pwd.yaml

Z4nzu:

The Nginx variable `PROXY_READ_TIMEOUT` is misspelled as `PROXY_READ_TIMOUT`, causing it to not update in the configuration file. This misspelling leads to an issue where the value is not being properly updated in the config file.

This PR will solve the issue and will update variable in configuration file correctly.

### The screenshot below illustrates that the value for `TIMEOUT` has not been updated in the configuration file, while the value for `client_max_body_size` has been successfully updated.
image
### Value provided via .env file
image

### DIFF
--- a/example.env
+++ b/example.env
@@ -38,7 +38,7 @@ UPSTREAM_REAL_IP_RECURSIVE=
 
 # All Values Allowed by nginx proxy_read_timeout are allowed, default value is 120s
 # Useful if you have longrunning print formats or slow loading sites
-PROXY_READ_TIMOUT=
+PROXY_READ_TIMEOUT=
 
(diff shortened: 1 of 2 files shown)


khoran:
This typo also exists in compose.yaml, which makes it impossible  to set this variable myself and have it propagate into the nginx.conf file:
`PROXY_READ_TIMEOUT: ${PROXY_READ_TIMEOUT:-120}`

If I spell it correctly, it doesn't get used by compose.yaml, but if I use the misspelling, it doesn't get used by the template for nginx.conf .

Issue #1148, opened 2023-06-11
upgrade images for dependencies


revant:
https://github.com/frappe/frappe_docker/actions/runs/5232576176/jobs/9447372249

- [x] Debian bullseye to bookworm
- [x] NodeJS v18
- [x] Python v3.11

revant:
https://github.com/frappe/frappe/issues/21326 update after this is complete

revant:
Note: Debian Bookworm has been released https://bits.debian.org/2023/06/bookworm-released.html. Images can start using `bookworm` instead of `bullseye`. May need changes in `apt-get` dependencies.

PR #1158, opened 2023-06-25
ci: refactor

revant:
use debian bookworm
use updated wkhtmltopdf
override python and nodejs versions


revant:
close #1148

surajshetty3416:
@revant [this got merged](https://github.com/frappe/frappe/issues/21326). You can proceed.

PR #1166, opened 2023-07-02
fix: set node version to 18 for frappe develop

netchampfaris:


revant:
Available in `frappe/bench:latest` now

https://github.com/frappe/frappe_docker/actions/runs/5439240544/jobs/9890976201

Issue #1175, opened 2023-07-10
Installer.sh script is broken: Venv cannot be found


cgallay:
## Description of the issue
Venv seems not to be present in the currently deployed version of [frappe/bench](https://hub.docker.com/r/frappe/bench).

I would recommend that some tags are pushed to the docker hub and not only the latest one as otherwise it's difficult to roll back to a working version when an issue arrives like right now.

## Steps to reproduce the issue

1. git clone frappe_docker
2. setup the dev container as in [Getting Started](https://github.com/frappe/frappe_docker/blob/main/docs/development.md)
3. configure any apps.json
4. Run `bash installer.sh` 

### Observed result
![image](https://github.com/frappe/frappe_docker/assets/3649143/bfc35e5e-9648-421a-8b6e-812a33fa3097)

### Expected result
Installation of the frappe apps.

### Stacktrace / full error message if available

```
Traceback (most recent call last):
File "/home/frappe/.bench/bench/commands/make.py", line 68, in init init
File "/home/frappe/.bench/bench/utils/render-py", line 105, in wrapper_fn return fn(*args, **kwargs)
File "/home/frappe/.bench/bench/utils/system.py", line 65, in init bench. setup. env (python=python)
File "/home/frappe/.bench/bench/utils/render-py", line 126, in wrapper_fn return fn(wargs, *kkwargs)
File "/home/frappe/.bench/bench/bench. py", line 362, in env self.run(f"(venv} env", cwd-self.bench. name)
File "/home/frappe/.bench/bench/bench.py", line 48, in run return exec_cmd (amd, cwd=cwd or self.cwd, _raise-_raise)
File "/home/frappe/.bench/bench/utils/_ init
-•py", line 154, in exec_amd
return_code = subprocess. call(spl_md, cwd=cwd, universal _newlines-True, env=env)
File "/home/frappe/-pyenv/versions/3.10.12/lib/python3.10/subprocess.py", line 345, in call with Popen (*popenargs, **kwargs) as p:
File "/home/frappe/-pyenv/versions/3.10.12/lib/python3.10/subprocess.py", line 971, in .
_init_
self._execute_child(args, executable, preexec_fn, close_fds,
File "/home/frappe/-pyenv/versions/3.10.12/lib/python3.10/subprocess.py", line 1863, in _execute_child raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: Errno 2] No such file or directory: "None*
ERROR: There was a problem while creating expo666
[...]

PR #1178, opened 2023-07-11
fix: use python for installer script

revant:
For #1175 

@cgallay does this work for you?


revant:
merging,

now this apps.json and the apps.json used in image build are same.

PR #1269, opened 2023-11-09
feat: add optional mailpit configuration

WebsiteDeveloper:


devcontainer-example/docker-compose.yml
```diff
@@ -20,6 +20,23 @@ services:
   #   volumes:
   #     - postgresql-data:/var/lib/postgresql/data
 
+  # Enable Mailpit if you need to test outgoing mail services
+  # See https://mailpit.axllent.org/
+  #  mailpit:
+  #    image: axllent/mailpit
+  #    container_name: mailpit
+  #    restart: always
```
revant, line 28: Don't set restart: always. Other services don't set it.

revant (changes requested):

revant:
Greenmail allows SMTP as well as IMAP locally:

```yaml
  greenmail:
    image: greenmail/standalone:2.0.0
    ports:
      - 3025:3025
      - 3110:3110
      - 3143:3143
      - 3465:3465
      - 3993:3993
      - 8181:8080
```

revant:
Either allow edit or remove the restart and container_name.

WebsiteDeveloper:
@revant sorry for the delayed response i removed the container name and restart.

Concerning greenmail, the main reason why we use mailpit is because it has a webinterface which greenmail does not to my knowledge

revant (approved):

revant:
> Concerning greenmail, the main reason why we use mailpit is because it has a webinterface which greenmail does not to my knowledge

It gives http rest api. You'll need to configure thunderbird or some mail client with `localhost:3025` for SMTP and `localhost:3143` for IMAP

PR #1283, opened 2023-12-06
Update troubleshoot.md

dntrply:
Added instructions for removal if want to start afresh.


revant:
Add it to the index at the top of the same page as well.

PR #1333, opened 2024-02-12
feat: add defaultInterpreterPath as a vscode setting as a starting point according to docs

WebsiteDeveloper:
See: https://github.com/microsoft/vscode-python/wiki/Setting-descriptions#pythondefaultinterpreterpath

This only adds a default settings.json file to local setup in container to make initial setup a bit more streamlined.


revant:
Follow https://github.com/frappe/frappe_docker/blob/main/CONTRIBUTING.md

PR #1345, opened 2024-03-01
Updated custom Containerfile to solve custom app download issue

malay-gondalia:
Updated custom Containerfile to solve custom app download issue as there was error downloading custom apps from user git

was unable to download apps from my git repo. added code to loop through app variable to download.

added code to download apps from user git if not found.


revant:
If you pass the PAT or credentials properly in apps.json it should work.

For your custom case you can always use your fork.

I managed to render apps.json with pat/creds using jinja and env var. If that is passed and is correct the it should work.

Try correct apps.json and try bench init manually on devcontainer

PR #1360, opened 2024-03-24
Install Latest Version to Mac/Linux Doc Update

arkenm:


revant:
I'll suggest just make it v15 instead of full semver.

arkenm:
Updated to your suggestion @revant, thank you.

AdiPathak97:
Hi, I am currently running the frappe docker on Ubuntu 24.04 LTS using the command in the readme, and it seems to work perfectly. May I know what difference it makes to mention platform in `pwd.yml`?

I would also like to know the reason behind making the version v15 instead of the specific version of the image.

Issue #1369, opened 2024-04-11
Documentation variables outdated


r1ckr:
## Description of the issue
In the documentation mentions the variable FRAPPE_VERSION, but in the example.env file it doesn't exists, instead it is ERPNEXT_VERSION.

I think the change is pretty simple, just change the variable in the example file to FRAPPE_VERSION.

revant:
Please send PR.

gparent:
My understanding is that `ERPNEXT_VERSION` is version of the ERPNext Docker image that is pulled when you are not building your own, whereas `FRAPPE_BRANCH` and `FRAPPE_PATH` are used for the Frappe framework repo and version. Is this correct?

r1ckr:
Then the documentation is the one that needs to be updated?

PR #1372, opened 2024-04-11
#1369 Update example.env

r1ckr:


revant:
This is ERPNext version ideally. The image is `frappe/erpnext`

search for `FRAPPE_VERSION` in the repo and you need to change it in docs md files, py scripts that run to auto bump env file.

PR #1376, opened 2024-04-17
feat: include HRMS in default image

ankush:
This is most common request for custom apps on docker images. While it's
possible to build images with custom apps it's still involved so we can
start with including top 1-5% apps at least.

https://github.com/frappe/frappe_docker/issues?q=is%3Aissue+hrms


ankush:
@WebsiteDeveloper yeah, maybe other apps can build some kind of pipeline on top of base images to provide it.

PR #1379, opened 2024-04-24
add build library for arm64 binaries

sudhegan:

added build library in custom image needed to build arm64 binaries which are not readily available 
eg: arm64 binaries need for print designer app refer this [link](https://github.com/frappe/print_designer?tab=readme-ov-file#faq)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,7 @@ venv
 
 # NodeJS
 node_modules
+
+#sudhegan frappe_docker git ignore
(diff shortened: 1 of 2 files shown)


revant:
is this required in base image? or you can move it to builder layer?

sudhegan:
actually i haven't tried it will  try and change that

sudhegan:
moved the library to base layer,pls ignore other commits

PR #1409, opened 2024-06-09
fix: make docs more explicit for beginners

Chev91:
As a novice, I played around with Play With Docker and had some success installing manually without Docker, but when it came to running Docker Compose locally on my own machine I got very lost. 

My proposal here is to add an additional step to help people in a similar position to make sure they can easily follow the quick start guide.


dnk8n:
I had similar issues, this would have saved me a few hours.

PR #1415, opened 2024-06-12
Update docker-compose.yml

Chev91:
A challenge I encountered while setting up the Development instance. The command 'bench init --skip-redis-config-generation frappe-bench' kept generating errors, which I discovered were due to an architecture mismatch between Docker and macOS. By updating the platform to linux/amd64 in my code, I was able to resolve this issue.

### DIFF
--- a/devcontainer-example/docker-compose.yml
+++ b/devcontainer-example/docker-compose.yml
@@ -2,6 +2,7 @@ version: "3.7"
 services:
   mariadb:
     image: docker.io/mariadb:10.6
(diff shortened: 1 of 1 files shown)


dnk8n:
This had me scratching my head for a while!

Definitely useful to be explicit here, as Frappe is not working as expected on alternative architectures.

FYI, I see the root directory `compose.yaml`  already has a similar change applied, would be especially great for newcomers to not hit this same issue.

PR #1452, opened 2024-08-07
doc: update docker compose command

lucca-rodrigues:
Updated command docker compose to docker-compose


samy4sam:
The docker-compose command is deprecated:
[https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose](https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose)

PR #1475, opened 2024-09-24
Update Traefik container from deprecated v2.6 to v2.11

gparent:

This updates a few yaml files to refer to an updated version of Traefik that is recently supported and fixes some recent vulnerabilities.

Fixes: https://www.cert.europa.eu/publications/security-advisories/2024-102/ and probably some others


gparent:
I do not have a big Frappe install but whatever testing I could get done did not trigger any issues with 2.11. Additionally, I could not find anything in the Traefik version migration documentation that suggested any breaking changes going from 2.6 to 2.11.

Issue #1487, opened 2024-10-06
YAML files cannot be generated anymore


tobehn:
Since this commit #1403 building the erpnext-one.yaml file is not possible anymore.

I always get the ERROR:
`invalid interpolation format for x-backend-defaults.image: "${ERPNEXT_VERSION:?No ERPNext version or tag set". You may need to escape any $ with another $.`

Even though i definitely have set ERPNEXT_VERSION the build process does not recognize it anymore. With the former image declaration everything works fine.
`  image: frappe/erpnext:${ERPNEXT_VERSION:?No ERPNext version set}`

with the new image declaration the generating the YAML fails and the file stay empty.
```
  image: ${CUSTOM_IMAGE:-frappe/erpnext}:${CUSTOM_TAG:-${ERPNEXT_VERSION:?No ERPNext version or tag set}}
  pull_policy: ${PULL_POLICY:-always}
```

I have also tried seting CUSTOM_IMAGE and CUSTOM_TAG to frappe/erpnext and the corresponding Version i wanted to use.

How do i need to adapt to this change?

revant:
Try 

```sh
--env-file /path/to/.env
```

tobehn:
Thank you very much for answering:
Well, thats exactly what i do.

```
docker compose --project-name erpnext-one \
  --env-file ~/gitops/erpnext-one.env \
  -f compose.yaml \
  -f overrides/compose.redis.yaml \
  -f overrides/compose.multi-bench.yaml \
  -f overrides/compose.multi-bench-ssl.yaml config > ~/gitops/erpnext-one.yaml
```

PR #1494, opened 2024-10-15
dollar escape for ERPNEXT_VERSION

tobehn:
I was not able to generate the `erpnext-one.yaml` and from `erpnex-one.env` in `~/gitops` anymore.
A detailed description of the failure can be found here: https://github.com/frappe/frappe_docker/issues/1487#issue-2568736420 

By adding the "$" escape in example.yaml the process ran flawlessly again.


tobehn:
I think i am wrong. This lead to using `image/erpnext` without image tag. So i ended up using version `16.0.0-dev`.

tobehn:
Its not doing what it was supposed to do

PR #1519, opened 2024-11-25
Changed pwd.yml

vogt31337:

changed some parameters, since syntax has changed:
- bench parameter for mariadb password is different
- mariadb parameter for password has changed

bench documentation (https://github.com/frappe/bench/wiki/Quick-Reference-Guide) shows the parameter for setting db password are different, from what is used here. Also with changing this, I don't get the warning/error that it cannot access the db anymore.

mariadb documentation (https://hub.docker.com/_/mariadb) states that the db password parameter is also different from what is used here. Sadly the env variables aren't easily documented therefore I'll just added the password parameter from what is described there.

### DIFF
--- a/pwd.yml
+++ b/pwd.yml
@@ -67,7 +67,7 @@ services:
           fi
         done;
         echo "sites/common_site_config.json found";
-        bench new-site --no-mariadb-socket --admin-password=admin --db-root-password=admin --install-app erpnext --set-default frontend;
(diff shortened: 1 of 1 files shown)


revant:
`--mariadb-root-*` is deprecated, `--db-root-*` is used now.

https://github.com/frappe/helm/issues/222

PR #1525, opened 2024-12-06
fix: --no-mariadb-socket is DEPRECATED

0xD0M1M0:
--no-mariadb-socket is DEPRECATED; use --mariadb-user-host-login-scope='%' (wildcard) or --mariadb-user-host-login-scope=<myhostscope>, instead.


revant:
@0xD0M1M0 can you also do that everywhere else? 

- pwd.yml
- bench easy-install.py
- helm create-site job

PR #1526, opened 2024-12-07
fix: --no-mariadb-socket & --mariadb-root-password is DEPRECATED

0xD0M1M0:
addition to #1525

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -71,7 +71,7 @@ def frappe_site(compose: Compose):
     site_name = "tests.localhost"
     compose.bench(
         "new-site",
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -96,13 +95,11 @@ def erpnext_site(compose: Compose):
     site_name = "test-erpnext-site.localhost"
     args = [
         "new-site",
+        # TODO: change to --mariadb-user-host-login-scope=%
```
revant, line 98: mentioned this

revant:
version-14 does not have that cli option, either send a PR to v14-hotfix or we'll remove it on release of v16

PR #1555, opened 2025-01-25
Remove redis-cache volume

sdfateh:
### Details of the Change:

This pull request addresses an issue where the assets.json file, now stored in Redis by Frappe, persists across new image builds due to the Redis cache volume being persistent. This prevents updates to assets.json from being reflected after a new build.

The proposed change removes the persistent volume for redis-cache, ensuring that the cache, including assets.json, is cleared during each new image deployment.

### Problem Solved:

By removing the persistent Redis volume, we prevent outdated assets.json data from being retained, ensuring the application always works with the latest assets after a build. This eliminates the need for manual intervention or additional automation to clear the cache.

Retaining the old assets.json file will cause the websites to miss certain assets:

image

### Why is This Necessary?

Without this change, developers or pipelines must manually clear the assets.json value from Redis or automate the process using additional scripts or hooks. Removing the persistent volume simplifies the process and avoids potential misconfigurations.

This PR was inspired by @ankush's comment on the issue https://github.com/frappe/frappe/issues/29901 , as he correctly pointed out that there is no need to persist in this volume.

### DIFF
--- a/docs/setup_for_linux_mac.md
+++ b/docs/setup_for_linux_mac.md
@@ -175,8 +175,6 @@ services:
     deploy:
       restart_policy:
         condition: on-failure
-    volumes:
-      - redis-cache-data:/data
 
   scheduler:
     image: frappe/erpnext:v15
@@ -207,7 +205,6 @@ services:
 volumes:
   db-data:
   redis-queue-data:
-  redis-cache-data:
   sites:
   logs:
 ```
--- a/overrides/compose.redis.yaml
+++ b/overrides/compose.redis.yaml
@@ -9,8 +9,6 @@ services:
 
   redis-cache:
     image: redis:6.2-alpine
-    volumes:
-      - redis-cache-data:/data
 
   redis-queue:
     image: redis:6.2-alpine
@@ -18,5 +16,4 @@ services:
       - redis-queue-data:/data
 
 volumes:
-  redis-cache-data:
   redis-queue-data:
--- a/overrides/compose.traefik-ssl.yaml
+++ b/overrides/compose.traefik-ssl.yaml
@@ -2,7 +2,7 @@ services:
   traefik:
     labels:
       # https-redirect middleware to redirect HTTP to HTTPS
-      # It can be re-used by other stacks in other Docker Compose files
+      # It can be reused by other stacks in other Docker Compose files
       - traefik.http.middlewares.https-redirect.redirectscheme.scheme=https
       - traefik.http.middlewares.https-redirect.redirectscheme.permanent=true
       # traefik-http to use the middleware to redirect to https
--- a/pwd.yml
+++ b/pwd.yml
@@ -176,8 +176,6 @@ services:
     deploy:
       restart_policy:
         condition: on-failure
-    volumes:
-      - redis-cache-data:/data
 
   scheduler:
     image: frappe/erpnext:v15.49.3
@@ -210,7 +208,6 @@ services:
 volumes:
   db-data:
   redis-queue-data:
-  redis-cache-data:
   sites:
   logs:
 


BelalShatara (approved):

gparent:
I hope this comment isn't too spammy - I just wanted to thank you for trying to improve on this and also to the others who posted solutions. I think you'll save many people a lot of frustration.

ankush (approved):
Change looks fine in principle. 

I'd let someone with better knowledge of frappe_docker merge it though.

sdfateh:
@Ph0en1XXX

 How did you handle volume removal? It seems like you deleted the sites volume

Ph0en1XXX:
I reinstalled the full images using the script for lms, but with my own image.

python3 ./easy-install.py deploy \
    --project=learning_prod_setup \
    --email=your_email.example.com \
    --image=ghcr.io/frappe/lms \
    --version=stable \
    --app=lms \
    --sitename subdomain.domain.tld

Ph0en1XXX:
> I did docker-compose down -v When reinstalledfull images using the script for lms, but with my own image.
> 
> python3 ./easy-install.py deploy --project=learning_prod_setup --email=your_email.example.com --image=ghcr.io/frappe/lms --version=stable --app=lms --sitename subdomain.domain.tld

Fixed with just using default image then changed to my own (in learning_prod_setup-compose.yml)

alexleach:
Hi, I've been dealing with my own upgrade issues (not related to this, however) when I found this PR.

I have suffered from this issue though, and had started removing the redis cache volume (as well as the assets volume, which I'd already been removing) manually on upgrades.

Anyway, I wanted to point out that removing the volume from the frappe_docker compose files doesn't stop a volume from being created, as it is still created by the redis base image:-

https://hub.docker.com/layers/library/redis/latest/images/sha256-aa4e560cd458a1a86f69730b334de33a01cb79ac737d537f829b46f0eda388b3

So, by removing the named volume, we end up with what I'll call volume bloat. Every time the images are upgraded, or containers recreated (`docker compose down && docker compose up -d`), we'll end up with persistent volumes with a random 64-character name in base64.

What would be ideal is if the cache and assets volumes were actually re-used and populated from scratch, as any other way adds maintenance burden when upgrading. I also always feel uncomfortable manually deleting volumes, as there's always a risk that some important volume like the database volume might be deleted, which would be a disaster!

aryacahil:
Is there a tutorial to resolve this? I'm sorry, I still don't understand and I'm a beginner.

aryacahil:
> Same question here, how to resolve this please

My suggestion is to just use Linux Ubuntu, bro. I switched from Docker to Linux Ubuntu and finally it worked.

hichem-code:
I use ubuntu and I Deploy with dokploy so i dont understand why

gparent:
You can use `docker volume prune` to easily remove the volumes.

You'll also want to regenerate your `compose.yaml` file after what Alex said.

developmentforpeople:
If make sense for anyone (@sdfateh , @revant , @ankush, @BelalShatara, ...) in my point of view, this not seems the proper solution for "solve assets issue".

If you use redis for store keys values coming from big computation methods (big queries or data processing and have hundreds of needed keys). You do not want a FLUSH ALL each deploy (that is the same effect).

Remember that this issue is related to assets, so make more sense add a process that clears that single key when new image starts:

```py
frappe.cache.delete_key(keys="assets_json", shared=True)
```

I will try to add it to the flow. More details of my answer here: https://github.com/frappe/frappe/issues/29901#issuecomment-3061334610

suhail135:
> docker exec -it ${PROJECT_NAME}-backend-1 bench --site "$site" clear-cache
> docker exec -it ${PROJECT_NAME}-backend-1 bench --site "$site" execute <path_to_before_migrate_hook>
> 
> docker exec -it ${PROJECT_NAME}-redis-queue-1 redis-cli FLUSHALL
> docker exec -it ${PROJECT_NAME}-redis-cache-1 redis-cli FLUSHALL

this also not wroking for me..

PR #1596, opened 2025-03-20
Create apps.json

wintogeno:


aldo-o:
I do no think it's a good idea to include other apps. Frappe Framework is different from ERPNext and other Frappe apps.

revant:
This just adds one apps.json file, It will trigger nothing on gh actions.

PR #1626, opened 2025-05-07
feat: add pdftk + print_designer + Chromium as pdf renderer

rasos:
fixes "pdftk test call failed" https://github.com/frappe/frappe/issues/27964#issuecomment-2802850845


revant:
Why added apps.json and base64 of apps.json?

rasos:
This was for our custom build with some additional apps. For a generic app including `print_designer` you would need to add at least the `print_designer` app, that is installing Chrome. 

We then build it with 

```
export APPS_JSON_BASE64=$(base64 -w 0 /path/to/apps.json)

docker build \
  --build-arg=FRAPPE_PATH=https://github.com/frappe/frappe \
  --build-arg=FRAPPE_BRANCH=version-15 \
  --build-arg=PYTHON_VERSION=3.12.8 \
  --build-arg=NODE_VERSION=22.14.0 \
  --build-arg=APPS_JSON_BASE64=$APPS_JSON_BASE64 \
  --tag=registry.osalliance.com/erpnext/erpnextcustom:v15.60.2 \
  --file=images/custom/Containerfile .
```

revant:
Remove if it was for you custom build, just send the relevant change

rasos:
Not being able to produce a PDF is a blocker to migrate from v14 to v15 not only to us, so that's why I think we should not wait until `print_designer` is integrated in frappe, as some issues are still pending. Another option would be to use [frappe_betterprint](https://github.com/neocode-it/frappe_betterprint) as its maintainer mentioned here https://github.com/frappe/frappe/issues/27965#issuecomment-2925327161 but that seems to be in an even more early stage.  Once one of them is integrated into frappe and fixes PDF handling, we can remove those apps from the build. 

Thanks for the feedback (it is part of our open source culture!). I changed the branch of  `print_designer` to main (I was always testing with develop before).  

No, `bench setup-chrome` is not necessary, as the `print_designer` is launching that by itself as soon as you activate the app (you need to do a `bench install-app` with any additional custom app from within the docker container or k8s pod): 

```
bench --site erp.mydomain.net install-app print_designer
Installing print_designer...
Chromium is not available. downloading...
Downloading Chromium from https://storage.googleapis.com/chrome-for-testing-public/133.0.6943.35/linux64/chrome-headless-shell-linux64.zip...
Downloading Chromium  [####################################]  100%          Extracting Chromium...
Renaming /home/frappe/frappe-bench/chromium/chrome-headless-shell-linux64 to /home/frappe/frappe-bench/chromium/chrome-linux
Making Chromium executable: /home/frappe/frappe-bench/chromium/chrome-linux/headless_shell
Chromium executable permissions set: /home/frappe/frappe-bench/chromium/chrome-linux/headless_shell
Chromium is ready to use at: /home/frappe/frappe-bench/chromium
```

maheini:
Thank you for the feedback - I appreciate it. 

Even though wkhtmltopdf has kind of a lot of issues, it's doing it's job in the end. I'd argue it's ok enough for most projects. But it's not on me to decide on that :smiley: 

Pay attention, bench install-app will install chromium, but only within the container it is executed in. If you got a multi-container setup (default compose.yaml), this won't install chromium within the scheduler/queue container.

Therefore, bench install-chrome should be run at build-time of the docker image. I already reported this issue in the print_designer repo [here](https://github.com/frappe/print_designer/issues/432) in order to raise awareness. (I suspect this might be also the case on your current production setup then)

I noticed something else, if I'm not mistaken: you added additional apps like webshop, ecommerce_integration, hrms and non_profit into apps.json, which wouldn't be installed by default and are unrelated to the print_designer specific implementation.

rasos:
I see, even when I install Chromium as a package here https://github.com/rasos/frappe_docker/blob/main/images/custom/Containerfile#L58 and thus is also available for the workers, the print_designer app expects it at some CHROMIUM_BINARY_PATH which is /home/frappe/frappe-bench/chromium/chrome-linux/headless_shell . So should the `bench setup-chrome`  just go after line https://github.com/rasos/frappe_docker/blob/main/images/production/Containerfile#L154 - maybe then we even do not need the` print_designer` app at all?

maheini:
`bench setup-chrome` stems from print_designer and won't be available without installing print_designer first. You need to install print_designer and it's direct dependencies and then add this command into the Dockerfile in order to get chromium installed at build-time:
```
# Bench commands won't work as root
USER frappe
RUN bench setup-chrome

# Maybe switch back to root if required...
# USER root
```

rasos:
When running ` bench setup-chrome` during build I am getting the error https://github.com/frappe/print_designer/issues/423#issuecomment-2952259809

> File "/home/frappe/frappe-bench/apps/frappe/frappe/init.py", line 391, in get_common_site_config
> common_site_config = os.path.join(sites_path, "common_site_config.json")
> ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> File "", line 76, in join
> TypeError: expected str, bytes or os.PathLike object, not NoneType
> ...
> RuntimeError: Failed to setup Chromium: expected str, bytes or os.PathLike object, not NoneType

and installing Chrome by doing a `bench install-app`  would require the `--site` parameter which we do not have in a generic image. 

So it seems not possible yet to create an image including a correct chrome setup for all workers.

maheini:
This is a bug in print_designer (afaik it's solved in the dev branch) https://github.com/frappe/print_designer/issues/437#issuecomment-2815906587

revant:
I'm closing this from here. 

We'll add it when frappe develop branch has it. 

Any other frappe official app, including print designer can have their github actions build it. 

CRM, HRMS, I helped configure. Rest of the new apps are following similar workflow.

rasos:
For those who can not wait for Chromium included in frappe, here is a `install_print_designer.sh`  script which I put in the `resources` directory that installs it for each site: 

```
#!/bin/bash
# Chromium gets lost when pod is restarted, so we install the print_designer app again
# Check if chromium directory is empty
if [ -z "$(ls -A /home/frappe/frappe-bench/chromium)" ]; then
    echo "Chromium directory is empty - installing print_designer on all sites"
    
    # Get all site directories (exclude 'assets' directory)
    sites=$(find /home/frappe/frappe-bench/sites -maxdepth 1 -mindepth 1 -type d -not -name 'assets' -not -name 'archived_sites')
    
    # Install print_designer on each site
    for site_path in $sites; do
        site=$(basename "$site_path")
        echo "Installing print_designer on site: $site"
        bench --site "$site" install-app print_designer --force
    done
else
    echo "Chromium directory not empty - skipping print_designer installation"
fi
```

In the `/images/custom/Containerfile` I added

```
# Copy print_designer app installation script and set permissions https://git.fairkom.net/hosting/k8s-deployments/erpnext/-/issues/7#note_105627
COPY resources/install_print_designer.sh /usr/local/bin/install_print_designer.sh
RUN chmod 755 /usr/local/bin/install_print_designer.sh
```

In the kubernetes deployment I added

```
# install again print designer in gunicorn if chromium got lost after pod restart
          lifecycle:
             postStart:
               exec:
                 command: ["/usr/local/bin/install_print_designer.sh"]
```

So even in case the pod gets moved to another worker node or restarted, it runs the script, and always installs the print_designer app (using --force option), which (re-)installs Chromium at every site into frappe/bench/chromium, as the frappe-bench directory is not persistent.

AbhishekJupalli-1409:
Actually i am trying to run it with the frappe docker github docker  pwd.yml its not working the script is being executed.
But when i exec in the frontend and backend container there is no folder of chromium

```
docker exec -it 348e1c2314ce bash
frappe@348e1c2314ce:~/frappe-bench$ ls
apps  config  env  logs  patches.txt  sites
docker exec -it 4da2ec7071d6 bash
frappe@4da2ec7071d6:~/frappe-bench$ ls
apps  config  env  logs  patches.txt  sites
```
the create site container logs
```

Installing print_designer...
Updating DocTypes for print_designer: [========================================] 100%
Chromium is not available. downloading...
Downloading Chromium from https://storage.googleapis.com/chrome-for-testing-public/133.0.6943.35/linux64/chrome-headless-shell-linux64.zip...
Downloading Chromium
Extracting Chromium...
Renaming /home/frappe/frappe-bench/chromium/chrome-headless-shell-linux64 to /home/frappe/frappe-bench/chromium/chrome-linux
Making Chromium executable: /home/frappe/frappe-bench/chromium/chrome-linux/headless_shell
Chromium executable permissions set: /home/frappe/frappe-bench/chromium/chrome-linux/headless_shell
Chromium is ready to use at: /home/frappe/frappe-bench/chromium
Chromium is already set up at /home/frappe/frappe-bench/chromium/chrome-linux/headless_shell
Updating Dashboard for print_designer
```
how can i fix this in docker compose.@rasos @revant

PR #1638, opened 2025-05-22
chore: Update EOL NodeJS to 20.19.2

gparent:

NodeJS 18 is EOL. This PR updates NodeJS to 20.

If a maintainer familiar with the build process could review that this does not break anything, it would be great. Seems ok here.


revant:
https://github.com/frappe/frappe_docker/blob/main/CONTRIBUTING.md#maintenance

This is a guide for upgrade. You need to change workflow as well. They'll be overridden in workflow making this change not useful.

gparent:
Thanks. I don't use the workflows so it worked for me.

PR #1642, opened 2025-06-02
chore: remove deprecated version directive

gparent:

This removes the deprecated `version` directive from compose files.

The `version` directive is deprecated as indicated by the couple error messages generated when using these files:

> WARN[0000] /home/erp/frappe_docker/overrides/compose.traefik.yaml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion

I'm submitting this mainly to clean up the output of the commands using these files and make it easier to see other more important warnings or error messages.

### DIFF
--- a/devcontainer-example/docker-compose.yml
+++ b/devcontainer-example/docker-compose.yml
@@ -1,4 +1,3 @@
-version: "3.7"
 services:
   mariadb:
     image: docker.io/mariadb:10.6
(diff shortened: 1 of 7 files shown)


gparent:
I don't think there's anything to explain- I can rebase or fix this if needed and I still think it is useful.

gparent:
Can this be reopened and merged? I don't understand what's wrong but I'm willing to fix it.

PR #1643, opened 2025-06-02
Add restart to backup cron

gparent:

Make sure backups don't stop happening following something like a docker update.

Makes it so the cron container keeps running after a restart.


gparent:
I don't think there's anything to explain- I can rebase or fix this if needed and I still think it is useful.

gparent:
Can this be reopened and merged? I don't understand what's wrong but I'm willing to fix it.

PR #1702, opened 2025-09-11
Use `docker compose` instead of `docker-compose`

meichthys:
Update to use new docker syntax now that `docker compose` has been standard for more than a few years.

### DIFF
--- a/docs/site-operations.md
+++ b/docs/site-operations.md
@@ -1,16 +1,16 @@
 # Site operations
 
-> 💡 You should setup `--project-name` option in `docker-compose` commands if you have non-standard project name.
+> 💡 You should setup `--project-name` option in `docker compose` commands if you have non-standard project name.
 
 ## Setup new site
 
 Note:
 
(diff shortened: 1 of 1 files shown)


ismxilxrif:
@meichthys @revant it seems the docs in this pr change is using `docker -p` instead of `docker compose -p`, which is incorrect, as the -p flag is only inside the compose-v2 plugin. same as with the `docker -p <project_name> exec <service_name>` commands, it should be `docker compose -p <project_name> exec <service_name>`

```
~/frappe_docker$ docker -p test compose logs
unknown shorthand flag: 'p' in -p

Usage:  docker [OPTIONS] COMMAND [ARG...]

Run 'docker --help' for more information
```

Cheers!

meichthys:
Oof, yes, that's a bit confusing! I'm sorry, I should have tested all of the commands.

I guess we should revert this PR for now.

meichthys:
See https://github.com/frappe/frappe_docker/pull/1708 which reverts these changes.

PR #1707, opened 2025-09-15
Improve devcontainer behaviour on ARM Macs

nliechti:
We discovered that the devcontainer experience is subobtimal on ARM Macs. I think these are very popular and maybe even Windows or Linux builds with ARM CPU's will someday be useful ;).

This change increased the performance of our local dev Setup significantly.

### DIFF
--- a/devcontainer-example/docker-compose.yml
+++ b/devcontainer-example/docker-compose.yml
@@ -42,8 +42,7 @@ services:
     image: docker.io/redis:alpine
 
   frappe:
-    image: docker.io/frappe/bench:latest
-    platform: linux/amd64
+    build: ../images/bench
     command: sleep infinity
     environment:
       - SHELL=/bin/bash
--- a/devcontainer-example/docker-compose.yml
+++ b/devcontainer-example/docker-compose.yml
@@ -2,7 +2,6 @@ version: "3.7"
 services:
   mariadb:
     image: docker.io/mariadb:10.6
-    platform: linux/amd64
     command:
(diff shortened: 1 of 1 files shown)


revant:
I'd suggest adding comments and documentation to do this change. Like how the postgres or mail mocking is done. Why force everyone to build image assuming everyone uses ARM64?

It saves time for many who can pull and use `frappe/bench:latest` image.

nliechti:
The build of the image takes about 20 Seconds or so on the first time you start the environment. This also protects from version divergence like in https://github.com/frappe/frappe_docker/pull/1706. I would propose to build it locally by default and make a comment to the prebuilt image.
But i can live with your suggestion as well.
The pinning to amd64 images can be removed though right?

revant:
Pinning to AMD64 is added by someone from community.

Line 45 the build line will take some time. There are apt get packages installed in Dockerfile. It will take more that 2 sec for no cache fresh installs depending on network.

Comment the build: line and add instructions in comments for usage like how pgsql and mock mail service is doing.

nliechti:
I changed it to use de prebuilt image again. I think we need to make sure the used Versions of Node, Python and so on are in sync in all used Images.

revant:
> I think we need to make sure the used Versions of Node, Python and so on are in sync in all used Images.

https://github.com/frappe/frappe_docker/blob/main/CONTRIBUTING.md#maintenance I added a note here for maintenance when versions are released.

PR #1714, opened 2025-09-23
Update single-server-example.md. added bench setup requirements after site install for proper documentation

malay-gondalia:
…l app dependencies from pyproject.toml or requirements.txt are installed.

added 
docker compose --project-name erpnext-one exec backend bench setup requirements

Running bench setup requirements after site creation ensures:

All app dependencies from pyproject.toml or requirements.txt are installed.

Any app hooks or modules that read common_site_config.json will work correctly.

You only need to run bench setup requirements once per bench, even if you create multiple sites later.


revant:
No need to do bench setup requirements if you've added dependencies to pyproject.toml or requirements.txt of frappe app.

This is incorrect as it'll corrupt the state of container in which this is executed.

PR #1751, opened 2025-11-20
Update ERPNext image version to v16.0.0-beta.1

alhraryahmed:

### DIFF
--- a/pwd.yml
+++ b/pwd.yml
@@ -2,7 +2,7 @@ version: "3"
 
 services:
   backend:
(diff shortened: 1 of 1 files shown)


DanielRadlAMR:
I will close this PR without merging because:
- There is no image with this tag yet, so it would not work as intended.
- v16 is still in beta and shouldn’t be the default selection until the full release.
- The tags don’t need to be updated manually — a workflow already handles tag updates.

PR #1756, opened 2025-11-28
Change port mapping from 8080 to 8081

alloujean:


alloujean:
8080 TO 8081

DanielRadlAMR:
Please provide more information on why this change would be useful or which bug it fixes. I don’t see the reason for changing the port.

I will close this if no further action is taken.

PR #1758, opened 2025-12-03
feat: publish version 16 for helm chart

revant:


DanielRadlAMR:
@revant 
image
Shouldn't this be set to `v16` as well?

PR #1777, opened 2025-12-23
docs: added deployment methode chooser

DanielRadlAMR:
Clarifies the different ways `frappe_docker` can be used and when each setup is appropriate.

Adds a new Getting Started document that distinguishes:

- `pwd.yml` (quick exploration)
- VS Code Devcontainers (local development)
- Easy Install script (automated production)
- `compose.yml` + overrides (manual production)

Also documents Docker immutability and why installing apps into running containers is not supported.

Docs-only change; no behavior or configuration changes.


DanielRadlAMR:
@asieftejani would you mind checking out my changes especially regarding the easy-install script 
https://github.com/AMR-Mannesmann/frappe_docker/blob/docs-01-getting-started/docs/01-getting-started/01-choosing-a-deployment-method.md#3-easy-install-script-from-frappebench

asieftejani:
@DanielRadlAMR thanks for following through with this - LGTM

PR #1796, opened 2026-01-16
chore: Update Traefik to version 3.6 and add migration doc

### LINKED ISSUE #1797: Upgrade Traefik from v2.11 to v3.6 because of EOL Feb 1, 2026
Currently Traefik in this Docker Frappe setup is used in release version **2.11**.
Security updates for this version will end on **February 1, 2026**. After that, no further updates will be published for the underlying image, so we need to upgrade to **Traefik v3.6** for both production and development setups.

Links/Reference:
[Traefik-Releases](https://doc.traefik.io/traefik/deprecation/releases/)
[EndOfLife](https://endoflife.date/traefik) 

Traefik v3 changed their rule syntax so it needs to be updated i added extra docs for migrations steps of already running instances

My Proposal and Related PR 
- #1796 
- #1798

Rocket-Quack:
Closes #1797

## Summary
Currently, Traefik is used in release version **2.11**. Security updates for this version will end on **February 1, 2026**.
This means that no updates for Traefik will be released for the underlying image, so it is necessary to update to version **3.6** for production and development setups. 

[Traefik-Releases](https://doc.traefik.io/traefik/deprecation/releases/)
[EndOfLife](https://endoflife.date/traefik)

In the new version, the rule syntax has been adjusted for the specification of domain sites and host names.
This updates the repo to use Traefik version 3.6 and also provides a doc for migration, and the previous docs have been adapted to the new syntax rule.

## Changes
- bump Traefik images to v3.6 in all proxy overrides
- migrate routing rules to v3 syntax (HostRegexp + Host OR rules)
- replace SITES with SITES_RULE in overrides, env and docs
- add Traefik v2→v3 migration guide for existing setups

## Breaking Changes
- `SITES` env variable replaced by `SITES_RULE` for consistent rule definition
- existing deployments must update env files and regenerate compose config for traefik and frappe-containers

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -31,13 +31,7 @@ def _add_sites_var(env_path: Path):
             "test-erpnext-site.localhost",
             "test-pg-site.localhost",
         )
-        sites_list = ",".join(f"`{site}`" for site in sites)
         sites_rule = " || ".join(f"Host(`{site}`)" for site in sites)
-        content = re.sub(
-            rf"SITES=.*",
-            f"SITES={sites_list}",
-            content,
-        )
         content = re.sub(rf"SITES_RULE=.*", f"SITES_RULE={sites_rule}", content)
         f.seek(0)
         f.truncate()


DanielRadlAMR:
First of all, thanks for this important contribution!

While this update should be merged to keep our dependencies up to date, I’m a bit concerned because it introduces breaking changes. Until now, we didn’t even have releases to mark such changes — I just created the very first release, which should allow us to handle breaking changes more safely.

I recommend splitting this PR into two separate ones:

### 1. First PR: Update Traefik to `v3.6` but keep backward compatibility by using:
```yaml
- traefik.http.routers.frontend-http.ruleSyntax=v2
```

### 2. Second PR: Fully migrate to all `v3` components. This PR would correspond with a new major release.

Rocket-Quack:
Thx :) I agree with you. 

Now that we also have release versions, the whole thing is easier to manage. I would also open a "develop" branch to better manage releases. In the current case, main can be retained for release versions, currently the 1.0.0, and branch develop would then be for introducing new changes, with a release merge back to main for updates. 

I would then first create a new PR so that we only specify the rule syntax for version 2 in the version 3 image, and then prepare the rest for a major update.

DanielRadlAMR:
Great!
Looking forward to your "updated" PRs.

You’re right about the two-branch setup, but it would require some additional workflows or workflow changes. Someone would need to build that—you could also open a third PR for it, of course 😄. Furthermore, since this is a bigger change to the repo’s structure, I wouldn’t merge it myself, as I’m “just” a collaborator. A proper maintainer should handle that.

DanielRadlAMR:
First part is merged, please resolve the mergeconflicts

DanielRadlAMR:
I think we should change `core.defaultRuleSyntax`, which you set in #1798, to `v3` as well — or remove it altogether. There’s no direct technical issue with the current PR since the router should overwrite this setting, but it’s a bit confusing to set different versions in different places and have them overwrite each other.

Rocket-Quack:
I agree to remove it complete from routers and cores. Since the rule syntax has now been set to Traefik version 3 anyway, which is also the recommended approach from [Traefik](https://doc.traefik.io/traefik/migrate/v2-to-v3/#final-configuration-cleanup).

DanielRadlAMR:
Why did you rename the `SITES` variable to `SITES_RULE`?

Does it somehow better align with Traefik syntax? Since this name can be chosen arbitrarily, I’m not sure why it needs to be changed otherwise.

Rocket-Quack:
I renamed `SITES `to `SITES_RULE` because this variable is more clearly recognizable as an actual Traefik router rule expression. What Traefik evaluates in the routers is no longer just a simple list of domains, but actually a rule syntax with options such as regex, for example. [Traefik Router Rule](https://doc.traefik.io/traefik/reference/routing-configuration/http/routing/router/#opt-rule) maybe interessting too: [Traefik Rules & Priorities](https://doc.traefik.io/traefik/reference/routing-configuration/http/routing/rules-and-priority/)
Because of that i think specifying `SITES_RULE` is better than implicitly treating the variable as `SITES` a “list of websites” and having a magic box generating the rules.

**TLDR** `SITES_RULE` immediately makes it clear that the variable expects a rule definition not just site names with comma seperation and matches Traefik rule terminology

PR #1798, opened 2026-01-19
chore: Transition preparations to update Traefik Image to version 3.6

Rocket-Quack:
# Transition Summary
This PR represents a kind of transition to prepare for the full update to **Traefik v3.6**. See #1796 , which contains the breaking changes and is considered a major update. A big thanks to @DanielRadlAMR 
This PR serves to update traefik images to version 3.6 and uses the rule syntax v2 for migration and further use of V2 rules via:
 `--core.defaultRuleSyntax=v2`

# Short problem description from main PR
Currently, Traefik is used in release version 2.11. Security updates for this version will end on February 1, 2026.
This means that no updates for Traefik will be released for the underlying image, so it is necessary to update to version 3.6 for production and development setups.

[Traefik-Releases](https://doc.traefik.io/traefik/deprecation/releases/)
[EndOfLife](https://endoflife.date/traefik)

# Changes:
- bump Traefik images to v3.6 in all proxy overrides
- Add note that `SITES` will soon be replaced by `SITES_RULE` in the major update

### DIFF
--- a/docs/02-setup/04-env-variables.md
+++ b/docs/02-setup/04-env-variables.md
@@ -60,9 +60,9 @@ SITES=`mysite.example.com`
 SITES=`{any:.+}`
 ```
 
->Note: `SITES` currently supplies the host list that goes into the `Host(...)` rules. Once the Traefik v3 transition is completed, the plan is to switch to `SITES_RULE` (full rule expression) and deprecate `SITES` to map the Traefik Rule Syntax. 
+> Note: `SITES` currently supplies the host list that goes into the `Host(...)` rules. Once the Traefik v3 transition is completed, the plan is to switch to `SITES_RULE` (full rule expression) and deprecate `SITES` to map the Traefik Rule Syntax.
 
->Currently the used syntax is pinned to v2 via `--core.defaultRuleSyntax=v2`).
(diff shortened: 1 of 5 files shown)


DanielRadlAMR:
Whats the difference between setting
```yaml
--core.defaultRuleSyntax=v2
```
in the proxy service and setting
```yaml
- traefik.http.routers.frontend-http.ruleSyntax=v2
```
in the frontend service?

i just found [this](https://doc.traefik.io/traefik/migrate/v2-to-v3-details/#remediation_18)

DanielRadlAMR:
Okay, after looking a bit further, this just sets the default value, which will be used if no further `ruleSyntax` label is set on the router

DanielRadlAMR:
@Rocket-Quack pls fix the spelling:
image

Rocket-Quack:
> Okay, after looking a bit further, this just sets the default value, which will be used if no further `ruleSyntax` label is set on the router

Yes I think that is better than including it with every router, for example, if different router definitions are made.

asieftejani:
@DanielRadlAMR @Rocket-Quack 
You might want to take a look at https://github.com/frappe/builder/issues/477#issuecomment-3852757962 
Not sure if its a ripple effect

PR #1800, opened 2026-01-29
feat: add nginx-proxy with acme companion as an alternative to traefik

Rocket-Quack:
Since I have also used nginx reverse proxy and acme companion in other projects, I think this is a good addition for frappe in docker as well. 
Many other setups already run with nginx‑proxy which means frappe_docker fits better into existing stacks.
You can choose what you want to use yourself. Traefik or, if necessary, nginx.

## Summary
Introduce an nginx-proxy + acme-companion alternative to Traefik

## TODOs (Open for discussion)
- [x] new docs: choosing a reverse proxy: Traefik vs nginx‑proxy
- [x] new docs: nginx‑proxy HTTPS setup example
- [x] new docs: complete setup instructions for using Traefik or nginx

## Changes
- Add compose overrides for nginx-proxy (HTTP) and nginx-proxy + acme-companion (HTTPS)
- Document the new overrides in setup docs and getting-started
- Split reverse-proxy variables in env docs (Traefik vs nginx-proxy)
- Add NGINX_PROXY_HOSTS examples and ACME/HTTP-01 notes

## Notes
- Traefik and nginx-proxy are mutually exclusive do not enable it both in the same stack.

### DIFF
--- a/docs/02-setup/08-single-server-nginxproxy-example.md
+++ b/docs/02-setup/08-single-server-nginxproxy-example.md
@@ -1,0 +1,177 @@
+# Single Server Example (nginx-proxy + acme-companion)
+
+This guide demonstrates a single-server setup using nginx-proxy and acme-companion for HTTPS. It is best for a small number of hostnames and a single bench. If you need multiple benches or advanced routing, use the Traefik-based example instead.
+
+We will setup the following:
+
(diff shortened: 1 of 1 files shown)


revant:
LGTM. Doesn't cause problem with existing setup. Merging.

PR #1802, opened 2026-01-30
Update Python and Node.js versions in Containerfile

### LINKED ISSUE #1801: Why not defaulting to python 3.14 and node 24 in custom image Dockerfile?
V16 needs python 3.14 and node 24, why not setting the default build ARGs accordingly in `images/custom/Dockerfile`

david-loe:
fixes #1801

### DIFF
--- a/images/custom/Containerfile
+++ b/images/custom/Containerfile
@@ -115,7 +115,7 @@ RUN if [ -n "${APPS_JSON_BASE64}" ]; then \
 
 USER frappe
 
-ARG FRAPPE_BRANCH=version-15
+ARG FRAPPE_BRANCH=version-16
 ARG FRAPPE_PATH=https://github.com/frappe/frappe
 RUN export APP_INSTALL_ARGS="" && \
   if [ -n "${APPS_JSON_BASE64}" ]; then \


DanielRadlAMR:
Please update this PR so that all image files share the same defaults

david-loe:
the github action to update the erpnext version should also be changed to version 16

DanielRadlAMR:
https://github.com/david-loe/frappe_docker/blob/patch-1/images/bench/Dockerfile

Line 130: The Node version should be updated to 24.13.0.

In this Dockerfile, I noticed that multiple Python and Node versions are installed to maintain backward compatibility.

Currently, some of the latest bench images have trouble running Frappe v15. We might want to address this to ensure v15 support.

My suggestion would be to drop v14 compatibility and add v15 instead. We could also rename the Python/Node version arguments to something more generic (e.g., `PYTHON_VERSION_PREV` / `NODE_VERSION_PREV`) rather than referencing a specific version.


You missed to update frappe branch in the custom dockerfile.

DanielRadlAMR:
Just noticed today is the official EOL of `version-14`. So dropping compatibility should be fine in my opinion.

revant:
> Just noticed today is the official EOL of `version-14`. So dropping compatibility should be fine in my opinion.

Yes we should. We have old versions on registry.

Should we tag frappe_docker repo or create a version-14 branch for legacy users?

DanielRadlAMR:
I’d prefer a tag over a `version-14` branch. This is a point-in-time snapshot rather than a separate development line. A clearly named tag (e.g. `v14-final` / `v14-eol`) would also make it obvious this is the last supported v14 state.

PR #1808, opened 2026-02-11
fix: enable pnpm via corepack for apps that require it (e.g. drive)

Tguntenaar:
The drive app's frontend build runs `npm run check-pnpm && pnpm install
&& pnpm build`. In minimal Docker images, the non-root frappe user
cannot reliably `npm install -g pnpm`, so drive's install-pnpm.sh
silently fails and bench init aborts with exit code 127 during asset
building. I've made a PR their too so it doesn't silently fail, however that doesn't fix that there is not access to pnpm.

Enabling corepack in the base stage (where we still run as root) makes
pnpm available system-wide without requiring a separate global install.

### DIFF
--- a/images/custom/Containerfile
+++ b/images/custom/Containerfile
@@ -46,6 +46,7 @@ RUN useradd -ms /bin/bash frappe \
     && nvm install ${NODE_VERSION} \
     && nvm use v${NODE_VERSION} \
     && npm install -g yarn \
+    && corepack enable pnpm \
     && nvm alias default v${NODE_VERSION} \
     && rm -rf ${NVM_DIR}/.cache \
     && echo 'export NVM_DIR="/home/frappe/.nvm"' >>/home/frappe/.bashrc \
--- a/images/production/Containerfile
+++ b/images/production/Containerfile
@@ -43,6 +43,7 @@ RUN useradd -ms /bin/bash frappe \
     && nvm install ${NODE_VERSION} \
     && nvm use v${NODE_VERSION} \
     && npm install -g yarn \
+    && corepack enable pnpm \
     && nvm alias default v${NODE_VERSION} \
     && rm -rf ${NVM_DIR}/.cache \
     && echo 'export NVM_DIR="/home/frappe/.nvm"' >>/home/frappe/.bashrc \


DanielRadlAMR:
why would you need `pnpm` in first place?

Tguntenaar:
Hi @DanielRadlAMR, thanks for your response. In this case its the frappe drive app that is using pnpm but I assume there would be others

https://github.com/frappe/drive/blob/7d22e5a9751a8781d90afc8a6f8e782ae505cbd4/package.json#L9

DanielRadlAMR:
Hmm, that’s interesting—do you know why they chose pnpm?

Most other Frappe apps use Yarn, so using pnpm would create duplicate structures, which isn’t ideal IMO.

Tguntenaar:
Some of the newer apps use `pnpm` apparently

https://github.com/frappe/drive/pull/594#issuecomment-3941095843

DanielRadlAMR:
@Rocket-Quack what do you think about this?

Rocket-Quack:
LGTM 🦆👍 especially the versions for pnpm which are defined by the apps themselves so the responsibility doesnt lie with us in the repo and we dont have to deal with any maintenance but still enable the user to use it

DanielRadlAMR:
Should be added to the production image as well as it it used in the layered image.

Rocket-Quack:
@DanielRadlAMR As I understand it other apps would only use it like **Drive**
Production image itself uses ERPNext and Frappe and has no custom apps that need this. 
But for the consistency, they could still implement it there as well.

DanielRadlAMR:
My point was that the production image is reused in the layered one, and the layered image allows installation of custom apps.

Maybe I’m missing something here?

Rocket-Quack:
Yeah you're right my mistake 🙈
It uses the base from production so we can add it to production too
Shouldn't be much of an extra work

DanielRadlAMR:
@Tguntenaar can you add the line to the production image as well?

Tguntenaar:
Alright!

Rocket-Quack:
LGTM! Merging

PR #1814, opened 2026-02-14
Create apps.json

raytrix71:


Rocket-Quack:
I think this PR was a mistake, as `apps.json` is not maintained in this repo but must be filled out by each user on the instances to be operated, depending on which apps are to be installed.

So I think this PR can be closed @raytrix71

PR #1815, opened 2026-02-16
feat: add chromium pdf generator support

trustedcomputer:
This PR adds support to the custom and production images for the new chromium PDF generator in version-16.

See https://github.com/frappe/frappe/pull/34191.

### DIFF
--- a/images/bench/Dockerfile
+++ b/images/bench/Dockerfile
@@ -23,6 +23,8 @@ RUN apt-get update \
     libharfbuzz0b \
     libpangoft2-1.0-0 \
     libpangocairo-1.0-0 \
+    #Chromium
+    chromium-headless-shell \
     # to work inside the container
     locales \
     build-essential \


DanielRadlAMR:
why did you use `chromium-headless-shell`?

frappe/press uses `chromium-browser`

https://github.com/frappe/press/pull/3517

trustedcomputer:
@DanielRadlAMR First, because this frappe feature came directly from the print_designer app, which always used chromium-headless-shell:

https://github.com/frappe/print_designer/blob/develop/print_designer/install.py#L210C1-L216C3

Second, chromium-browser is the full browser with a GUI, which is unnecessary bloat (558MB vs 470MB in my testing on Debian Trixie, an 18.7% increase).

Third, because that's the better fit: https://developer.chrome.com/blog/chrome-headless-shell:

>  The old Headless mode [implemented by chromium-headless-shell] is a lightweight wrapper around Chromium's //content module, and it therefore has substantially fewer dependencies. Specifically, it does not require X11/Wayland, D-Bus, and is in some ways more performant than the fully-fledged Chrome browser. This makes it suitable for use cases such as automated screenshotting or web scraping.
>     New Headless [implemented by chromium-browser] on the other hand is the real Chrome browser, and is thus more authentic, reliable, and offers more features. This makes it more suitable for high-accuracy end-to-end web app testing or browser extension testing.
>     In other words, there's a trade-off between performance and authenticity. Which Headless mode is most fitting for you? It depends on your use case.

Having said all that, there aren't any technical reasons why the full chromium package would not work, but do you see any technical benefits to installing the full browser for PDF generation? I can't think of any.

DanielRadlAMR:
Thanks for the followup!

This should be updated as well:
https://github.com/frappe/frappe_docker/blob/main/images/bench/Dockerfile

DanielRadlAMR:
Will merge this after the `bench` Dockerfile is updated as well!

trustedcomputer:
> Will merge this after the `bench` Dockerfile is updated as well!

Thanks, @DanielRadlAMR , I've added the commit for bench dockerfile, tests are running.

PR #1817, opened 2026-02-21
fix(nginx): forward correct proto/origin for socket.io behind proxy

0samashaikh:
When running behind Traefik v3.6 with HTTPS, `$scheme` inside nginx is always `http` because Traefik terminates TLS. This caused the `Origin` header forwarded to the Socket.IO server to be `http://domain.com` instead of `https://domain.com`, resulting in a mismatch that silently dropped all real-time events to the browser.

Fixed by using `X-Forwarded-Proto` (set by Traefik) to correctly determine the actual client scheme when constructing the `Origin` header.

### DIFF
--- a/resources/core/nginx/nginx-template.conf
+++ b/resources/core/nginx/nginx-template.conf
@@ -6,6 +6,12 @@ upstream socketio-server {
 	server ${SOCKETIO} fail_timeout=0;
 }
 
+# Parse the X-Forwarded-Proto header - if set - defaulting to $scheme.
+map $http_x_forwarded_proto $proxy_x_forwarded_proto {
+	default $scheme;
+	https https;
+}
+
 server {
 	listen 8080;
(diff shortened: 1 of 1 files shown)


DanielRadlAMR:
Add a description

Rocket-Quack:
I tested this locally and on two single stack and multi stack setups and everything looks fine. 
It also fixes the issues from #1819 based from my tests

But i would also add to the `location @webserver`:
`proxy_set_header X-Forwarded-Proto $proxy_x_forwarded_proto;`
instead of
`proxy_set_header X-Forwarded-Proto $scheme;`

Because the same TLS-termination mismatch can affect normal requests too and so we keep the backend behavior consistent

trustedcomputer:
I can confirm this fixes #1819 for me in develop, version-16, and version-15.

revant:
Merging because it fixes socketio for a users facing the issue.

For webserver we can create another pr instead of blocking this.

Issue #1829, opened 2026-03-04
allow to use db user apart from db SA account


mastertIT:
Allow the ability to use a DB user apart from the SA root mariadb user.

this allows the container to utilize a server on the network without requiring the root SA account (which is a security risk)

DanielRadlAMR:
First of all, just to clarify:
Sites themselves do not use the root user—they use a site-specific user.

Only bench uses the root user, for example, when creating a new site (i.e., creating a new database).

Therefore, I think the actual security risk here is minimal.

Your usecase is still problematic, but from v16 onwards, you can specify a different DB user with the `--db-user` flag in bench operations.

mastertIT:
Is there the ability to declare the db user as an environment variable or not yet?

DanielRadlAMR:
I’m not sure I fully understand what you mean. From what I understood, you’re not running the DB in a container but instead using an already existing one on the network. Why would you want to set the user and password as environment variables?

You can just use the flag mentioned in the bench commands.

mastertIT:
correct, running mariadb on the network elsewhere.
however the ability to push the db user and password into the container on a 'compose up'  would allow the ability to deploy a new container without additional post config.

from a corp network perspective that's pretty standard. i.e have a db user that has permissions to create tables that isn't the SA root.

DanielRadlAMR:
As mentioned, a user is created during site creation.
The root user shouldn't be needed from that point on.

mastertIT:
Still an issue when the root sa is shared with other databases...
Why is there not a way to manually create the user with the correct db permissions and set it as the softwares default db user...

I don't understand why basic role based access control at the backend level isn't taken seriously by the new era of software development

DanielRadlAMR:
as already mentioned the site uses its own user, NOT the root user

adithyaappu:
There are site specific users. The username and password of of a site specific DB user is usually stored in the site config JSON. This user will not have access to anything else than the respective site. 

Now your question may have a point, if you mean why root user is needed for a specific application. In other words if you mean is it possible to create a non-root user, who will have higher permissions like create db, and use it as the common-admin/default user of Frappe. 

I'm not really sure if this is possible, as the document says root user, and I use docker based deployment, so never checked it. But theoretically it may be possible to have a non-root limited user which kind of act as the common-admin/default user. If you have setup for this, can you test and say the if it works?

mastertIT:
[...]

PR #1831, opened 2026-03-04
Feat/add postgres demo

nishanthabimanyu:
### Summary
This PR adds a **PostgreSQL quick-start demo** (`pwd-postgres.yml`) to the `frappe_docker` ecosystem. Currently, `pwd.yml` only supports MariaDB by default, which creates a high barrier for users who prefer PostgreSQL to evaluate the stack.

### Technical Details
- **Syncs with v16:** Uses `postgres:15-alpine` as the base image.
- **Automated Configuration:** Sets `DB_TYPE=postgres`, `DB_PORT=5432`, and updates the `create-site` command to use `--db-type=postgres` and the correct root credentials.
- **Verification:** Tested locally  confirmed `bench new-site` successfully initializes on PostgreSQL.

### Why this is needed
- **Relates to #1829:** Addresses the community's demand for better database flexibility.
- **Fixes User Friction:** Solves the manual effort described in historical issues like **#19994 (Frappe)** where users struggled to adapt the `pwd.yml` for Postgres.
- **One-Click Experience:** Provides parity between MariaDB and PostgreSQL for the "Play with Docker" (PWD) workflow.

### Checklist
- [x] Verified with local Docker logs

### DIFF
--- a/pwd-postgres.yml
+++ b/pwd-postgres.yml
@@ -1,0 +1,216 @@
+services:
+  backend:
+    image: frappe/erpnext:v16.7.3
+    networks:
+      - frappe_network
+    deploy:
+      restart_policy:
+        condition: on-failure
+    volumes:
+      - sites:/home/frappe/frappe-bench/sites
+      - logs:/home/frappe/frappe-bench/logs
+    environment:
+      DB_HOST: db
+      DB_PORT: "5432"
+
+  configurator:
+    image: frappe/erpnext:v16.7.3
+    networks:
+      - frappe_network
+    deploy:
+      restart_policy:
+        condition: none
+    entrypoint:
+      - bash
+      - -c
+    command:
+      - >
+        ls -1 apps > sites/apps.txt;
+        bench set-config -g db_type postgres;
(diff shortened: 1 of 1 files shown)


DanielRadlAMR:
How is this related to #1829?

The issue suggests avoiding the root DB user, but this PR introduces a new Compose file for a completely different database.

Also, I don’t think adding another pwd file is a good idea. pwd is intended mainly for exploration, and for that use case the specific database shouldn’t really matter.

nishanthabimanyu:
Thank you for the feedback, @DanielRadlAMR. You're absolutely right—I over-reached by
 linking this to #1829, and I understand the desire to keep the root directory lean.
 
 I'll close this PR. However, I noticed that overrides/compose.postgres.yaml is currently
 using an outdated Postgres version (13.5) and is missing the DB_TYPE environment variables
 required for a seamless v16 setup. Would you be open to a PR that updates and fixes that
 existing override instead?

DanielRadlAMR:
For sure! PRs that fix issues are always welcome!

PR #1834, opened 2026-03-06
fix(overrides): update PostgreSQL to v15 and improve v16 compatibility

nishanthabimanyu:
This PR updates the existing PostgreSQL override to align with Frappe v16 standards.

- Upgrades Postgres from 13.5 to 15-alpine.
- Adds a healthcheck (pg_isready) for more reliable startup.
- Automates the configurator to set db_type: postgres, removing the need for manual configuration steps.

### DIFF
--- a/docs/06-migration/03-postgres-major-version-upgrade.md
+++ b/docs/06-migration/03-postgres-major-version-upgrade.md
@@ -6,29 +6,34 @@ Upgrading PostgreSQL from version 13 to 15 is a major version jump. Since Postgr
 
 1. **Backup Existing Data (Version 13):**
    Before updating your compose file, ensure your containers are running and perform a dump of all databases.
+
    ```bash
    docker exec -it <project_name>-db-1 pg_dumpall -U postgres > full_dump.sql
    ```
 
 2. **Stop and Remove Containers:**
+
    ```bash
    docker compose down
    ```
 
 3. **Delete Old Data Volume:**
    PostgreSQL 15 cannot read data created by version 13. You must remove the existing volume (Warning: this deletes the old data directory).
+
    ```bash
    docker volume rm <project_name>_db-data
    ```
 
 4. **Update Image and Start (Version 15):**
    Update your `overrides/compose.postgres.yaml` (or pull the latest changes) and start the containers.
+
    ```bash
    docker compose up -d
    ```
 
 5. **Restore Data:**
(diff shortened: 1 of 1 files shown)


nishanthabimanyu:
I've simplified the override to match the project's existing MariaDB patterns, updated to Postgres
15-alpine, and added a healthcheck for better reliability. This should be a clean, update
to the existing infrastructure

DanielRadlAMR:
Is there a reason to use the Alpine image here? trixie is the default base.
If not, I’d suggest sticking with the defaults and using `postgres:15.17`.

Did you test in-place updates?

nishanthabimanyu:
@DanielRadlAMR  -- I chose the Alpine image primarily for the smaller footprint (~190MB vs ~450MB) and faster pull times, which I thought aligned with the 'quick setup' nature of the override files. However, I completely understand the preference for consistency with the project's trixie base and will gladly switch it to `postgres:15.17` to match.

Regarding in-place updates -->  I have not tested a 13 to 15 migration path. Since this is a major version jump, I assume users with existing data would need to perform a `pg_dump` or use a migration tool. Since this override is often used for new evaluation environments, I prioritized the v16-compatible version 15, but I'm open to suggestions if we should provide a migration warning or stay on a lower version for compatibility.

DanielRadlAMR:
I would stick to the default just because it is the default (there is usually a reason for that). If you think the pros of the Alpine version are significant and useful for this project, you can certainly argue for it. Since I am not using PostgreSQL myself, I will not comment further on that part — I just wanted to make sure there is some intention behind this selection.

For MariaDB, you can simply set the environment variable `MARIADB_AUTO_UPGRADE=1`, but it seems PostgreSQL upgrades work differently.

We should provide a migration guide though.

nishanthabimanyu:
Hey @DanielRadlAMR, I’ve updated the PR to match exactly what we discussed!

1. **Switched to `postgres:15.17`**: I’ve moved away from the Alpine image to stick with the project's Debian defaults as you suggested.
2. **Added the Migration Guide**: Since Postgres doesn't have a simple 'auto-upgrade' flag like MariaDB, I’ve added a dedicated guide for the v13 to v15 jump to help existing users.
3. **Updated Docs**: I also updated the Site Operations guide to reference the new migration docs and the simplified setup.

 Let me know if this covers everything you were looking for! Thanks again for the help.

PR #1837, opened 2026-03-10
fix(overrides): remove redis services from mariadb network in multi-b…

IB2357:
…ench setup

In a multi-bench setup with a shared `mariadb-network`, services with the same name from different projects (e.g. `redis-cache`, `redis-queue`) can conflict. Docker DNS will randomly (round-robin) resolve these service names across different projects, so a bench can end up connecting to a different bench’s Redis. This resulted in problems like `frappe.boot.assets_json` returning asset hashes from the wrong bench.

This change removes `mariadb-network` from `redis-cache` and `redis-queue`, keeping them only on the internal `bench-network` so each bench always talks to its own Redis.

### DIFF
--- a/overrides/compose.multi-bench.yaml
+++ b/overrides/compose.multi-bench.yaml
@@ -37,12 +37,10 @@ services:
   redis-cache:
     networks:
       - bench-network
-      - mariadb-network
 
   redis-queue:
     networks:
       - bench-network
-      - mariadb-network
 
 networks:
   traefik-public:


DanielRadlAMR:
@Rocket-Quack whats your take on this?

Rocket-Quack:
From the currently supported override combinations I do not see another official setup that requires bench internal Redis to be reachable over the `mariadb-network`
The only thing this could break is a custom setup where someone accesses a bench Redis from outside via the shared DB network but I would not consider that a supported dependency because its not a part of the official multi-bench topology.

So I think the change is valid as is and fixes the issue with multi-bench instances using a shared MaraDB
Merging it 👍

PR #1838, opened 2026-03-11
Ensure security headers are preserved in /files location

RinZ27:
Existing Nginx configuration for the `/files/` location block uses `add_header Content-disposition "attachment"`, which inadvertently clears all security headers defined at the `server` block level due to Nginx's inheritance model.

This change explicitly re-adds the critical security headers (HSTS, X-Frame-Options, X-Content-Type-Options, etc.) to the `/files/` location block. Additionally, the `always` flag is added to ensure these headers are sent even for non-2xx/3xx response codes.

This ensures that uploaded files (such as .html or .svg) are still protected by established security policies, preventing potential Clickjacking and XSS risks.

### DIFF
--- a/tests/test_frappe_docker.py
+++ b/tests/test_frappe_docker.py
@@ -85,7 +85,9 @@ def test_files_html_security_headers(
     frappe_site: str, tmp_path: Path, compose: Compose
 ):
     file_path = tmp_path / "testfile.html"
-    file_path.write_text("<html><body>This is a Frappe Docker test html file</body></html>")
+    file_path.write_text(
+        "<html><body>This is a Frappe Docker test html file</body></html>"
+    )
 
     compose(
         "cp",


Rocket-Quack:
@RinZ27 thanks for pointing that out :)
But I would prefer not to duplicate the same header set in two places.

A cleaner approach would be to move the shared headers into a small nginx snippet and then include it both at `server` and in the `/files/` while keeping `Content-Disposition` local to `/files/`. Can you refactor this into using a nginx include?

RinZ27:
@Rocket-Quack refactored the common security headers into a snippet at `resources/core/nginx/security_headers.conf` and included it in both the `server` and the `/files/` location block. Kept `Content-Disposition` local to the `/files/` block as requested.

Rocket-Quack:
Great Work @RinZ27 I pushed a few follow-up commits to make the fix safe to merge.
Besides moving the shared headers into a snippet also I updated the container files and the nginx include path. 
The original snippet path reflected the repository layout but nginx resolves against the filesystem inside the built container. 
Without copying the snippet into the image and referencing it through a real runtime path nginx would not be able to find it.

I also added regression coverage for `/files/` security headers.

RinZ27:
Thanks @Rocket-Quack for refining the nginx paths and adding the regression tests. The shared snippet approach is definitely cleaner.

PR #1852, opened 2026-03-30
fix: nginx redirects leak internal port

### LINKED ISSUE #1851: Trailing slash redirects leak internal port 8080 behind Proxy
When accessing the Frappe app behind Traefik with HTTPS enforced, URLs ending with a trailing slash (e.g., `https://my.erp.com/app/`) are redirected to the internal container port (`:8080`, e.g. `https://my.erp.com:8080/app`).

### Proposed fix :

1. Add `port_in_redirect off;` in the nginx server block.
2. Optionally, modify trailing-slash rewrites to use `$proxy_x_forwarded_proto://$host` instead of default `$scheme`.

DanielRadlAMR:
fixes #1851 

When accessing the Frappe app behind Traefik with HTTPS enforced, URLs ending with a trailing slash (e.g., `/app/`) are redirected to the internal container port (`:8080`).

Setting `port_in_redirect off;` fixes this.
This setting tells nginx never to append the listening port to redirects it generates.

### DIFF
--- a/resources/core/nginx/nginx-template.conf
+++ b/resources/core/nginx/nginx-template.conf
@@ -15,7 +15,7 @@ map $http_x_forwarded_proto $proxy_x_forwarded_proto {
 server {
 	listen 8080;
 	server_name ${FRAPPE_SITE_NAME_HEADER};
-  port_in_redirect off;
+  absolute_redirect off;
 	root /home/frappe/frappe-bench/sites;
 
 	proxy_buffer_size 128k;


Rocket-Quack:
That's actually a good solution the only issues I can think of are setups with customized ports and those behind a proxy that want to access the container directly via that port for example in local development or direct access via localhost:8080 because redirects then no longer properly account for the explicitly specified port.

`absolute_redirect off` might be a better solution in my opinion to keep local setups working and to prevent port leaks.

DanielRadlAMR:
I somehow thought `port_in_redirect` only affects redirects when no port is explicitly set.

I’ll test it with `absolute_redirect off` instead and merge if it works on my side.

Thanks for the feedback!

PR #1858, opened 2026-04-03
docs: fix dead link for environment variables reference.

sujaldev:


Rocket-Quack:
When providing the technical reference instead of linking to the GitHub Page (which i know is generated from the docs) 
i would recommend linking directly to the Markdown file in the repository, so that people who clone the project and may not have direct access to the pages it can find the Markdown file within the project itself

Here is the link to the md file use that please: https://github.com/frappe/frappe_docker/blob/main/docs/02-setup/04-env-variables.md

This would then also apply to other documentary links in the project

sujaldev:
Makes sense. I've made the requested changes.

Issue #1860, opened 2026-04-05
fix(security): APPS_JSON_BASE64 build-arg leaks private repo tokens in image layer metadata


OmarElaraby26:
## Description of the issue

When building custom images with private apps, `apps.json` contains GitHub PATs embedded in repository URLs. The current Containerfiles (`images/custom/Containerfile` and `images/layered/Containerfile`) pass this as a Docker build argument (`--build-arg=APPS_JSON_BASE64`). Docker build arguments are **permanently recorded in image layer metadata**, meaning anyone with access to the image can trivially extract private repo tokens.

This is a known Docker anti-pattern documented by Docker themselves: [SecretsUsedInArgOrEnv](https://docs.docker.com/reference/build-checks/secrets-used-in-arg-or-env/).

**Impact:** Any image built with private apps using the documented workflow has credentials permanently embedded. Images pushed to registries (even private ones) expose PATs to all users with pull access.

**Affected files:**
- `images/custom/Containerfile`
- `images/layered/Containerfile`
- `docs/02-setup/02-build-setup.md`
- `docs/02-setup/08-single-server-nginxproxy-example.md`

## Context information (for bug reports)

- **frappe_docker:** main branch (latest as of 2026-04-05)
- **Docker Engine:** any version with BuildKit support
- **Host OS:** any (issue is in Containerfile / docs, not host-specific)

## Steps to reproduce the issue

1. Create `apps.json` with a private repo URL containing a PAT:
   ```json
   [
     {
       "url": "https://USER:ghp_XXXX@github.com/org/private-app.git",
       "branch": "main"
     }
   ]
   ```
2. Build the custom image following the current docs:
   ```bash
   docker build \
     --build-arg=APPS_JSON_BASE64=$(base64 -w 0 apps.json) \
     --build-arg=FRAPPE_PATH=https://github.com/frappe/frappe \
     --build-arg=FRAPPE_BRANCH=version-16 \
     --tag=custom:latest \
     --file=images/custom/Containerfile .
   ```
3. Inspect the image:
   ```bash
   docker image history --no-trunc custom:latest
   ```

### Observed result

The full base64-encoded `apps.json` (including PATs) is visible in the `ARG` layer metadata. Decoding it reveals the plain-text GitHub PAT:

```bash
echo "<base64_from_history>" | base64 -d
# {"url": "https://USER:ghp_XXXX@github.com/org/private-app.git", ...}
```

### Expected result

[...]

PR #1863, opened 2026-04-08
feat: Easy-Docker TUI

Rocket-Quack:
## Easy Docker
Hello everyone 🦆 after a somewhat longer development period, I’d like to introduce you to **easy-docker**. 

**easy-docker** is a gum-powered Terminal UI (TUI) for creating and managing Frappe Docker stacks. It is designed to reduce the amount of manual Docker Compose, image build, and Bench work needed to get started, especially for users who are new to this setup.

@DanielRadlAMR already had the chance to check out a few teasers, but I think the time is right now, and the project has reached a stage where we can let the community take a look at what's been developed so far and get your feedback on it.

It’s designed as a tool to enable even beginners with no prior experience in Docker or Bench to run their desired apps.

There are many more screenshots and step-by-step instructions but this should already provide a solid first impression. 
If you’re interested i would be glad to hear your feedback and invite you to try it out.

Main_Menu_Overview

stack_creation_app_selection

Instead of assembling the workflow manually **easy-docker** guides you through the stack lifecycle. It currently covers:  
- stack creation
- app and branch selection
- image preparation and build
- stack runtime actions
- site management actions such as create, migrate, backup, and delete

## What is included

The current branch includes:

- guided stack creation for `single-host`
- guided setup flow for `split-services`
- persistent stack state under `.easy-docker/stacks/<stack>`
- app selection and branch selection flows
- custom image build flow
- stack runtime actions:
  - start
  - restart
  - stop
  - delete
  - status
- single-host site management flows:
  - create site
  - install app
  - uninstall app
  - migrate
  - backup
  - delete site

## Risk

This is still a large feature so there may still be uncovered edge cases in manual end-to-end flows

The main persistent contract introduced by this PR is the stack state written under:
- `.easy-docker/stacks/<stack>/metadata.json`
- `.easy-docker/stacks/<stack>/apps.json`
- `.easy-docker/stacks/<stack>/<stack>.env`

## TODOs

- [x] Expand documentation
- [x] Support split topology (proxy, database etc.)
- [x] Import for existing stacks (if applicable) (Will be later shipped)
- [x] Automated GitHub Actions for version synchronization across branches (Kicked because of YAGNI)
- [x] Test setup for validation of future updates
- [x] The changes to the terms of use regarding the listing of apps, as proposed in #1860 

I'll provide more information as time goes on, and as soon as I'm satisfied with the final solution of each step.

## Why?

I hope this helps a lot of people, and since we’ve also seen that we have to wait a very long time for changes to **easy-install.py**, for example, and really have no control over the situation in case of breaking changes.

> 
>The whole process is designed so that all files are created as if you were executing the necessary commands manually, meaning >that, if needed, the setup can be adjusted manually or further developed without having to use EasyDocker. My goal is therefore to ensure that anyone can run and use Frappe apps on their own with the appropriate tools.

I'm always happy for your feedback or suggestions 😄


DanielRadlAMR:
Hey, just a heads up — I won’t be able to look into this for the next two weeks, sorry :(

Rocket-Quack:
> Hey, just a heads up — I won’t be able to look into this for the next two weeks, sorry :(

No Problem better save and slow than fast with many bugs 😄

ab22593k:
Trying to test it on fedora 43 got: 
$ ./easy-docker.sh
at 11:21:31 tput: unknown terminal "xterm-256color"

Rocket-Quack:
Thanks for reporting @ab22593k  
This was caused by failing to resolve the terminal capabilities on that Fedora 43 environment. 
easy-docker was already ignoring the exit code but it still let the warning leak to stderr during startup.

I have updated the terminal handling so the missing `tput` are now handled quietly.
I also added a regression test for this case and expanded CI to run the easy-docker Bats suite across Ubuntu, Debian, and Fedora 43 so types of this issues are caught earlier.

Where you using a minimalistic fedora or an special container image so that i can easyier reproduce that usage case?

But it should work now on Fedora if you want to test that again 🫡

ab22593k:
Great job @Rocket-Quack ; thanks

AliShaatani:
great job we looking for it

DanielRadlAMR:
@Rocket-Quack, is it possible to add a custom app, for example via a GitHub link?

Rocket-Quack:
The GitHub Repo shouldn't be a problem since Bench supports it by default.
I had something like that implemented before but I didnt quite like the look and feel of it.
I have to figure out how to add the look and feel for custom apps or someones own repo-apps using a PAT token which would also be helpful. 

But in short yes of course 😄

DanielRadlAMR:
I think this would be very useful to implement.

By the way, sorry for my late response. I read through the documentation, and it looks good to me. Unfortunately, I do not think I will have enough time to review this in greater depth for the rest of the year.

That said, this PR should not get stuck simply because I am unable to take a closer look. Since this introduces an additional deployment method, I would suggest merging it and collecting broader feedback from the community.

There are two things you may want to consider before merging:

1. Add this deployment method to the documentation:
   https://github.com/frappe/frappe_docker/blob/main/docs/01-getting-started/01-choosing-a-deployment-method.md

2. Consider noting that it is new and still under development—perhaps describing it as an alpha release. Of course, this only makes sense if that accurately reflects its current state. If you are confident that it works reliably, including in edge cases, such a note could be misleading.

However, I would not start deprecating the other deployment methods, such as the easy-install script, just yet. I think we should first collect feedback from the community before taking steps in that direction.

I would also suggest announcing this in a post on the Frappe Forum. :)

Really great work!

Rocket-Quack:
No problem 👍
I haven't had much time myself but I try to incorporate a few more changes that I have noticed. 
As soon as that's ready I will merge the whole thing. 

I see it exactly the same way you do I still need to include a few new features that have been added to the Frappe Docker repo that I haven't covered yet.

PR #1866, opened 2026-04-10
chore(vscode): exclude build artifacts and deps from file watcher

DanielRadlAMR:
### Reduce VS Code inotify watches for Frappe bench workspaces
Adds `files.watcherExclude` entries to prevent the "*Unable to watch for file changes (ENOSPC)*" error when developing in a Frappe bench workspace.

The excluded folders do not need to be watched IMO.

Also excludes `__pycache__/` and `*.pyc` from both the file watcher and the Explorer, as these are auto-generated Python bytecode files that are never useful to browse or track.


DanielRadlAMR:
@Rocket-Quack if you don't mind having a quick look at this - if you don't see a reason to keep watching these files either, I'd merge it 🙂

Rocket-Quack:
I'm on your side :) 
LGMT!

PR #1873, opened 2026-04-15
chore(vscode): show git folder

DanielRadlAMR:
Addition to #1866 

Although the `.git` folder is hidden by default, it can be valuable to inspect during development.
This change makes it visible in the explorer for convenience.


Rocket-Quack (approved):
LGTM

PR #1875, opened 2026-04-15
fix: cap setuptools below v82 to retain pkg_resources

UmakanthKaspa:
**Problem**
setuptools v82 removed `pkg_resources`. Frappe v15 depends on it via `google-api-python-client`, causing all requests to fail: **AttributeError: 'NotFoundPage' object has no attribute 'app_path'**

 **Fix**
Cap setuptools in the site venv after `bench init`, using the same constraint already merged in frappe/bench#1699.

### DIFF
--- a/images/custom/Containerfile
+++ b/images/custom/Containerfile
@@ -137,7 +137,8 @@ RUN export APP_INSTALL_ARGS="" && \
     /home/frappe/frappe-bench && \
   cd /home/frappe/frappe-bench && \
   echo "{}" > sites/common_site_config.json && \
-  find apps -mindepth 1 -path "*/.git" | xargs rm -fr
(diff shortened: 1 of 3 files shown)


DanielRadlAMR:
Hey! Thanks for that PR.

I don’t think this is something we should fix here. If this is needed at runtime, the dependency should be defined somewhere else - like in the Frappe repo directly, for example.

The best long-term solution would be to remove all `pkg_resources` dependencies.

How do you know `google-api-python-client` is the problem? I checked it as well and didn’t find any `pkg_resources` dependencies there.

UmakanthKaspa:
Thanks for the review! Will open an issue in frappe/frappe instead — closing this for now.

DanielRadlAMR:
You’re very welcome!

This might be interesting for you: https://github.com/frappe/frappe/issues/38145

PR #1879, opened 2026-04-20
Fix/remove nested sites assets volume

### LINKED ISSUE #1850: Asset mismatch between backend and frontend due to sites/assets volume declaration
## Description of the issue

When using the official production/custom images from frappe_docker, the container declares both:
```
/home/frappe/frappe-bench/sites
/home/frappe/frappe-bench/sites/assets
```
as Docker volumes.

If docker-compose only mounts sites (as in the standard setup), Docker creates anonymous per-container volumes for sites/assets.

This can lead to a split asset state between containers, causing:
	•	mismatched assets.json
	•	missing hashed CSS/JS bundles
	•	partially broken ERPNext UI

**Symptoms**

	•	Browser console shows 404 errors for hashed assets (e.g. desk.bundle.*.css)
	•	MIME type errors (text/html instead of text/css)
	•	ERPNext UI partially broken (layout/styles missing)
	•	Custom apps (e.g. Mint) may still appear to work
	•	Running bench build in one container makes things worse
	•	Restarting containers does not fix the issue

**Root Cause**

The image defines:
```
VOLUME [
  "/home/frappe/frappe-bench/sites",
  "/home/frappe/frappe-bench/sites/assets",
  "/home/frappe/frappe-bench/logs"
]
```
But the default compose setup only mounts:

`sites:/home/frappe/frappe-bench/sites`

As a result:
	•	/sites → shared named volume ✅
	•	/sites/assets → anonymous volume per container ❌

This causes backend and frontend containers to read/write different asset directories, leading to manifest/hash mismatches.

1.	Have been using a  standard frappe_docker setup
2.     Build a new layered docker container to be able to include apps (docs/02-setup/02-build-setup.md )
3.	Install a custom app (e.g. Mint) or run bench build
4.	Access ERPNext UI
5.	Observe missing/broken assets in browser

You can verify the issue with:

`docker inspect <container> --format '{{json .Mounts}}'`

You will see different anonymous volumes mounted at:

`/home/frappe/frappe-bench/sites/assets`

### Observed result

Missing/broken assets in browser

### Expected result

Working front end

### Fix / Workaround

Explicitly mount a shared named volume for sites/assets across all relevant services:

docker-compose.override.yaml:

```
services:
  backend:
    volumes:
      - sites:/home/frappe/frappe-bench/sites
      - sites-assets:/home/frappe/frappe-bench/sites/assets

  frontend:
    volumes:
      - sites:/home/frappe/frappe-bench/sites
      - sites-assets:/home/frappe/frappe-bench/sites/assets

  configurator:
    volumes:
      - sites:/home/frappe/frappe-bench/sites
      - sites-assets:/home/frappe/frappe-bench/sites/assets

volumes:
  sites:
  sites-assets:
```

**Expected Behavior**

After recreating containers, backend and frontend will share the same asset state and the issue is resolved.

All containers should operate on a single coherent asset directory, so that:
	•	assets.json
	•	built JS/CSS bundles

are consistent across backend and frontend.

**Notes**

	•	This issue may not appear in simple vanilla (Production) setups but becomes visible when:
	•	installing custom apps
	•	rebuilding assets
	•	running multiple containers with different lifecycle timing
	•	The problem is subtle and difficult to diagnose without inspecting Docker mounts

ews-pgasser:
## What does this PR do?

Removes the nested `sites/assets` volume declaration from the layered `Containerfile` and adds documentation explaining the change and upgrade steps.

## Problem

Previous images declared `VOLUME /home/frappe/frappe-bench/sites/assets` separately alongside the `sites` volume. This created an implicit nested mountpoint inside the `sites` volume, which caused Docker to attach different anonymous volumes per container in multi-container setups — leading to inconsistent asset serving.

## Changes

- **`images/layered/Containerfile`**: Removed the redundant `sites/assets` volume declaration. `sites` is now the single shared mount, consistent with the compose setup.
- **`docs/02-setup/02-build-setup.md`**: Added an upgrade section explaining the breaking change and the required steps for users pulling the updated image.

## Upgrade steps for existing users

- Recreate all containers with `docker compose up --force-recreate`
- No `bench build` needed — this only fixes mount consistency, not the asset workflow

Fixes #1850

### DIFF
--- a/docs/02-setup/02-build-setup.md
+++ b/docs/02-setup/02-build-setup.md
@@ -74,20 +74,20 @@ podman build \
 
 ## Build args
 
-| Arg                  | Purpose                                                                                       |
-| -------------------- | --------------------------------------------------------------------------------------------- |
-| **Frappe Framework** |                                                                                               |
-| FRAPPE_PATH          | Repository URL for Frappe framework source code. Defaults to https://github.com/frappe/frappe |
-| FRAPPE_BRANCH        | Branch to use for Frappe framework. Defaults to version-15                                    |
-| **Custom Apps**      |                                                                                               |
-| (secret) apps_json   | Passed via `--secret=id=apps_json,src=apps.json`. Never use `--build-arg` for this file.      |
-| **Dependencies**     |                                                                                               |
-| PYTHON_VERSION       | Python version for the base image                                                             |
-| NODE_VERSION         | Node.js version                                                                               |
-| WKHTMLTOPDF_VERSION  | wkhtmltopdf version                                                                           |
-| **bench only**       |                                                                                               |
-| DEBIAN_BASE          | Debian base version for the bench image, defaults to `bookworm`                               |
-| WKHTMLTOPDF_DISTRO   | use the specified distro for debian package. Default is `bookworm`                            |
+| Arg                  | Purpose                                                                                         |
+| -------------------- | ----------------------------------------------------------------------------------------------- |
+| **Frappe Framework** |                                                                                                 |
+| FRAPPE_PATH          | Repository URL for Frappe framework source code. Defaults to <https://github.com/frappe/frappe> |
+| FRAPPE_BRANCH        | Branch to use for Frappe framework. Defaults to version-15                                      |
+| **Custom Apps**      |                                                                                                 |
+| (secret) apps_json   | Passed via `--secret=id=apps_json,src=apps.json`. Never use `--build-arg` for this file.        |
+| **Dependencies**     |                                                                                                 |
+| PYTHON_VERSION       | Python version for the base image                                                               |
+| NODE_VERSION         | Node.js version                                                                                 |
+| WKHTMLTOPDF_VERSION  | wkhtmltopdf version                                                                             |
+| **bench only**       |                                                                                                 |
+| DEBIAN_BASE          | Debian base version for the bench image, defaults to `bookworm`                                 |
+| WKHTMLTOPDF_DISTRO   | use the specified distro for debian package. Default is `bookworm`                              |
 
 # env file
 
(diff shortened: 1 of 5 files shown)


docs/02-setup/02-build-setup.md
```diff
@@ -130,6 +130,17 @@ This generates `compose.custom.yaml`, which you'll use to start all containers.
 
 > **NOTE**: podman compose is just a wrapper, it uses docker-compose if it is available or podman-compose if not. podman-compose have an issue reading .env files ([Issue](https://github.com/containers/podman-compose/issues/475)) and might create an issue when running the containers.
 
+# Upgrading from images with a nested sites/assets volume
```
Rocket-Quack: I think this upgrade note would fit better under the migration documentation rather than in the general build/setup docs

DanielRadlAMR:
can you add the fix to all relevant containerfiles?

Rocket-Quack (changes requested):

docs/02-setup/02-build-setup.md
DanielRadlAMR: nice catch

vkctl:
The migration note says "Recreate all containers (docker compose up --force-recreate). No bench build is needed." This is accurate for local Docker volumes, but breaks silently for setups using a persistent or NFS-backed sites volume.

What was happening before this PR:
Because sites/assets was declared as a separate VOLUME, Docker seeded a fresh anonymous volume for that path from the image on every new container. Persistent sites volumes never touched sites/assets/,  assets always came fresh from the image on each deployment.

What happens after this PR for persistent volumes:
sites/assets/ is now a subdirectory inside the persistent sites volume. The volume retains old assets across deployments. The fresh assets compiled into the image during bench init are invisible at runtime because the persistent volume mount shadows them. --force-recreate alone does not help.

What is actually required:
After deploying a new image, bench build --production must be run explicitly inside the backend container to write fresh assets into the persistent volume.

Suggest adding this to the migration note:
If your sites volume is persistent (NFS, named volume with retained data, or any external driver), run bench build --production in the backend container after recreating containers. The --force-recreate step alone is not sufficient in this case.

DanielRadlAMR:
@orey-aajaamu thanks for your input. We’re already working on this — you can check the issue above for details.

trustedcomputer:
Is bringing the project up with "--force-recreate" after new images a one-time migration step, or will this be necessary after every new image is built, assuming there is a change in assets?

siddharthaithena:
I have the same question as @trustedcomputer .

DanielRadlAMR:
Only needs to be done once

Issue #1828, opened 2026-03-03
Rework and centralize app image build and publish workflows for registries


Rocket-Quack:
Hello everyone 🦆, 

I am currently working on simplifying the operation of Docker for newbies or people who don't have much Docker experience with a script in the style of easy-install from Frappe using TUI ([gum](https://github.com/charmbracelet/gum)). 
What I noticed during development and for some time now is that the workflows for building a Docker image differ depending on the repo and apparently also use different approaches. 

I would like to tackle this and build the whole thing in a more systematic way. 

I have already given it some thought, but so far I haven't come up with a solution that I think is great and should be used.

> As a basis, I'm just going to take the four “big” Frappe apps (Erpnext, crm, helpdesk, and lms) and warn you that this could be a long text, but I think it's especially valuable for the future of Frappe apps and Docker.

First, I'd like to discuss the current situation.

## Current Situation

The current setup appears to be historically split:
ERPNext triggers image builds centrally in `frappe_docker`, while `CRM`, `Helpdesk` and `LMS` still build in their own repositories (using `frappe_docker` layered build context) and publish mainly to **GHCR**. 

This mixed model likely grew over time and is now harder to maintain consistently and is also difficult to understand from a technical perspective for Docker operation (except, of course, for people who build their own Docker image).

## Proposed future

I propose a hybrid model for (now the big four/in the future all) the Frappe Apps.
Whats a Hybrid Model?
ERPNext, CRM, HRMS and LMS should only define **when** an image build is triggered and **which app ref** should be built, while `frappe_docker` should own the reusable workflow for **how** images are built, tested, tagged, and published. (ERPNext is doeing something like that already)

### Practice example

In practice an app repo calls the reusable workflow in `frappe_docker` via `workflow_call`, passing inputs like `app_repo`, `app_ref`, and `frappe_branch`. The central workflow then builds the image using shared Dockerfiles from `frappe_docker` (typically `images/layered/Containerfile`), and pushes the resulting image to both registries (Docker Hub and GHCR app namespace, e.g. `ghcr.io/frappe/crm`).

```mermaid
flowchart TD
  A[frappe/crm release or dispatch] --> B[crm caller workflow]
  C[frappe/erpnext release or dispatch] --> D[erpnext caller workflow]
  B --> E[Checkout frappe_docker build context]
  D --> E
  E --> F[Build image]
  F --> H[(Docker Hub)]
  F --> I[(GHCR)]
```

[...]

PR #1881, opened 2026-04-21
feat: rework workflows and new image building workflow

Rocket-Quack:
Follow up to #1828

## Summary

This PR restructures the image workflow setup to make the core image pipeline easier to follow and to introduce a reusable workflow for downstream app images.

It keeps the existing core image behavior intact in principle but splits responsibilities more clearly:
- core image build and test
- core image publish
- downstream app image build

In addition, it extends the shared image publishing flow so that `base` and `build` are pushed to GHCR alongside the existing Docker Hub flow.

### Core workflow split

The previous core image pipeline was split into separate reusable workflows:

- `core-build-test-images.yml`
- `core-publish-images.yml`

And the orchestration workflows were updated accordingly:

- `core-build-develop.yml`
- `core-build-stable.yml`

The result is:
- version resolution, build, and test happen first
- publish happens only after successful test completion

### Stable version handoff between test and publish

The resolved `frappe` and `erpnext` tags are now passed forward explicitly as workflow outputs.

This avoids re-resolving versions during publish and ensures that the publish step uses the exact same tags that were tested.

### GHCR support for shared core images

The core image pipeline now also publishes:
- `base`
- `build`

to GHCR.

Docker Hub publishing remains in place, while GHCR is added as an additional destination for the shared foundation images.

### Reusable downstream app image workflow

This PR adds a new reusable workflow:

- `app-build-image.yml`

It allows another repository to call into `frappe_docker` to build a final app image from:
- an app repository
- an app ref
- an existing shared `base` / `build` image line
- a caller-defined target image name and registry

This workflow builds `images/layered/Containerfile`, installs the requested app and can then push the resulting app image to Docker Hub or GHCR

### Workflow naming cleanup

Workflow file names and visible workflow names were aligned to a clearer convention:
- `core-*` for shared image workflows
- `app-*` for reusable downstream app workflows
- `docs-*` for documentation publishing

This improves readability in both the repository and the Actions UI.

### Documentation

A new reference document was added to explain:
- the current workflow layout
- the current image flow
- naming conventions
- how downstream repositories can call the reusable app-image workflow

## Why

This change is intended to support a cleaner shared-image model for downstream Frappe app repositories.

The main goals are:
- make the core workflow structure easier to understand and maintain
- publish reusable `base` and `build` images to GHCR
- allow downstream repositories to build app images using a reusable workflow
- keep the tested and published core image versions aligned

## Compatibility

The core build behavior remains aligned with the previous setup:
- shared core images are still built and tested before publish
- Docker Hub publishing remains in place
- `base` and `build` gain GHCR publishing in addition to existing behavior

For `images/layered/Containerfile`, the default image source remains compatible with the current Docker Hub-based flow unless a different prefix is explicitly provided by the app-image workflow.

## Validated on fork
- core test workflows executed successfully
- publish stage was reached during fork validation
- temporary fork-specific publish gating used only for validation and reverted before the final PR state

## Follow-up

A downstream repository can now add a caller workflow that uses `frappe/frappe_docker/.github/workflows/app-build-image.yml` to build and publish its own app image based on the shared `base` and `build` images.


.github/workflows/core-build-stable.yml
DanielRadlAMR, line 1: Other Frappe repositories are currently using this workflow. For example:
https://github.com/frappe/frappe/blob/develop/.github/workflows/on_release.yml
https://github.com/frappe/erpnext/blob/develop/.github/workflows/docker-release.yml

Renaming our workflow will eventually break those workflows as well.

I think we should keep the old path available while internally switching to the new workflow. In parallel, we can identify which Frappe repositories still reference the old path and submit PRs to update them. Once all workflows have been migrated, we can safely remove the old path.

DanielRadlAMR (changes requested):

.github/workflows/core-build-stable.yml
Rocket-Quack: Added a temporary compatibility wrapper for build_stable.yml so existing release dispatches from frappe and erpnext should continue to work and uses the new intern workflow. Added also one for build_develop.yml just in case it is used.

What do you think of that solution i think its better than keeping old files as a double?

DanielRadlAMR: Haha yeah, that’s exactly what I expected you’d do 😄

LGTM!

DanielRadlAMR (approved):
LGTM

gparent:
I hope this isn't seen as spam but I would like to thank the people who have recently contributed to modernizing and making this great pipeline and Docker set up even greater. I bet you're making hundreds of lives easier downstream. I know I've certainly benefited from many of these changes. Building on the shoulder of giants and making it even better.

PR #1882, opened 2026-04-21
Use `mariadb-admin` for health check

simonsan:
mysql_upgrade is not existing anymore in the new 11.x images

### DIFF
--- a/pwd.yml
+++ b/pwd.yml
@@ -82,9 +82,11 @@ services:
     networks:
       - frappe_network
     healthcheck:
-      test: mariadb-admin ping -h localhost --password=admin
-      interval: 1s
-      retries: 20
+      test: ["CMD", "healthcheck.sh", "--connect", "--innodb_initialized"]
+      start_period: 5s
+      interval: 5s
+      timeout: 5s
+      retries: 5
     deploy:
       restart_policy:
         condition: on-failure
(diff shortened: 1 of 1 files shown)


DanielRadlAMR:
Check how the healthcheck is implemented in the override here:
https://github.com/frappe/frappe_docker/blob/main/overrides/compose.mariadb.yaml

It uses the built-in `healthcheck.sh` script

I suggest we follow the same approach here.

ews-pgasser:
> Check how the healthcheck is implemented in the override here: https://github.com/frappe/frappe_docker/blob/main/overrides/compose.mariadb.yaml
> 
> It uses the built-in `healthcheck.sh` script
> 
> I suggest we follow the same approach here.

@DanielRadlAMR I agree, we should stick to a standardized way (which happens to be the `healthcheck.sh`).
@simonsan It is a great catch, just change it similar to how [override/compose.mariadb.yaml](https://github.com/frappe/frappe_docker/blob/main/overrides/compose.mariadb.yaml) looks.

simonsan:
I PRed it because it was not working out of the box. Clone the repo, run with podman, get 500 due to create-site blocking. So this fixed it for me. 😅

simonsan:
Ah, got you. Thanks for the fixup. 👍🏽

PR #1888, opened 2026-04-23
fix(compose): add update-assets service

### LINKED ISSUE #1883: Asset mismatch between backend and frontend after running `bench migrate`
## Description of the issue

I run a CI powered, docker based setup with a custom image.

In my CI environment, I build the custom container image. After the image is built, I run some docker compose commands on my server:

```bash
docker compose down
docker compose pull
docker compose up -d
docker compose exec backend bench migrate
docker compose exec redis-cache redis-cli FLUSHALL
```

This worked well, until #1879. Now, the `bench migrate` command creates a new `assets.json`, which is stored with the frontend container as it is stored in `/home/frappe/frappe-bench/sites/assets/` folder, and with that in the `sites` volume.

The css/js files itself are stored in `/home/frappe/frappe-bench/apps/`, which is not shared between the containers. There are symlinks from the assets forlder to the apps folder. So after running the migration in the backend container, the css/js files in the backend container are updated, but not that files in the frontend container. But as the `assets.json` file is shared, the frontend cannot find the assets and the user get's 404 errors.

In the past, it looks like it worked by accident. I run `bench migrate` in the backend container, and bench updated css/js in the backend container, and the `assets.json` in a unnamed volume, which was not shared with the frontend container. So the frontend container still served the "old" version of these files.

I understand that the css/js files should not be build in productive environment. But I need to run the migrations.

How to run migrations without building the assets?

1. Build a custom app image
2. Update container using that image
3. Run `bench migrate` in the backend container

### Observed result

`assets.json` and accessable files are out of sync.

### Expected result

`bench migrate` should not manipulate `assets.json` file.

DanielRadlAMR:
fixes #1883

This PR implements the changes suggested by @ews-pgasser.

Recently, `sites/assets` was removed as a nested volume. This means `sites/assets` now fully belongs to the already mounted `sites` folder. As a result, assets are no longer updated correctly when new images are introduced.

This PR introduces a new service that runs once (similar to the configurator) and updates the assets accordingly.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -10,6 +10,8 @@ x-depends-on-configurator: &depends_on_configurator
   depends_on:
     configurator:
       condition: service_completed_successfully
+    update-assets:
+      condition: service_completed_successfully
 
 x-backend-defaults: &backend_defaults
   <<: [*depends_on_configurator, *customizable_image]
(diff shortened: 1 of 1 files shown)


DanielRadlAMR:
https://github.com/frappe/frappe_docker/issues/1883#issuecomment-4318460877

PR #1889, opened 2026-04-24
feat: add dynamic resolution for FRAPPE_PATH in Containerfile

ismxilxrif:
adds a dynamic resolution for folks that are using a private repo for the frappe app, and are using tokens to authenticate

### DIFF
--- a/images/custom/Containerfile
+++ b/images/custom/Containerfile
@@ -119,13 +119,19 @@ USER frappe
 ARG FRAPPE_BRANCH=version-16
 ARG FRAPPE_PATH=https://github.com/frappe/frappe
 RUN --mount=type=secret,id=apps_json,target=/opt/frappe/apps.json,uid=1000,gid=1000 \
+    --mount=type=secret,id=frappe_path,target=/tmp/frappe_path,uid=1000,gid=1000 \
   export APP_INSTALL_ARGS="" && \
+  if [ -f /tmp/frappe_path ] && [ -s /tmp/frappe_path ]; then \
+    RESOLVED_PATH=$(cat /tmp/frappe_path); \
+  else \
+    RESOLVED_PATH=${FRAPPE_PATH}; \
+  fi && \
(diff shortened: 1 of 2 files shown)


ews-pgasser:
Thanks for the PR! Just to understand the use case better: are you building custom images with a private Frappe fork?

Looking at the Containerfile, `apps.json` is already mounted as a secret in the same `RUN` step. You can put your private Frappe fork URL directly in there:

```json
[
  {
    "url": "https://token@github.com/org/private-frappe",  "branch": "version-16"
  }
]
```

This already solves the problem without any changes to the Containerfile, the secret mount ensures the token never appears in `docker history` or the final image.

If you are building in CI, most people run the build directly inside the GitHub Actions pipeline of the private repo, where the built-in `GITHUB_TOKEN` handles authentication transparently with no extra setup needed.

Could you describe your exact setup? It would help us understand if there is a genuine gap that `apps.json` doesn't already cover.

ismxilxrif:
@ews-pgasser  from what we're experiencing, is that bench init does not read the frappe repo in apps.json, but from the flags `--frappe-path` and `--frappe-branch`

jslocomotor:
Perhaps you can elaborate on how you are building images and deploying the stack?
Because if you're building a custom image, apps are already integrated in that image.
And we can confirm, that the build process with apps.json is working as is.

Or do you mean the Frappe framework? Are you trying to use a private fork of the framework inside apps.json for building?

ismxilxrif:
yes, we're using a custom fork of frappe, which we did try to embed inside apps.json, but bench init does not use that, but FRAPPE_PATH and FRAPPE_BRANCH instead

we have multiple private forks of frappe across v12 until v15

DanielRadlAMR:
Hmm, that’s interesting. Do you know what would happen if you removed `--frappe-branch=${FRAPPE_BRANCH}` and `--frappe-path=${FRAPPE_PATH}` and installed Frappe via `apps.json` as well?

Issue #1843, opened 2026-03-19
[RFC] Improving Documentation


adithyaappu:
# Background
During the discussion about ERPNext documentation, #1793 I just linked to a documentation framework called [Diataxis](https://diataxis.fr/start-here/). While completing the VitePress based Documentation site, I happen to notice that the documentation is actually in a poor form, and it is very difficult for a new user to learn and get started with Containerised Frappe.

# The Proposal
This is a request for comments, regarding restructuring and rewriting the documentation to make it more user friendly and easy to explore. A two phase approach is proposed where the first phase is adapting a better organisation of pages, and the second phase will be rewriting documentation pages to improve quality. Below I'm adding a new proposed structure for docs based on the Diataxis framework I mentioned above. 

## Points to note
1. Some pages are renamed to clearly convey what it contains. Current names of these pages, in a compressed form, are given in the parenthesis along with new name.
2. Some pages are combined together because the content is mostly same or very close. These are marked by specifying a `+` between the page names. 
3. The following list mainly shows the compressed name of the name of the page will be shown in the sidebar of VitePress Documentation. Which means more detailed titles can be specified in the page. 

# Proposed Structure

```
# Explanations and Concepts
1. Overview
2. Choosing a Method
3. Container Immutability
4. Frappe Concepts
5. FAQ

# Guides and Manuals
1. Quick Start on Unix Systems
2. Local Development Setup
3. Easy Install Script
4. Build Images (Build Setup)
5. Initialising Site (Start Container) + Site Operations
6. Creating Backups
7. Multi Bench Setup (Single Server Setup) + NGINX and ACME Companion
8. Caddy With HTTPS
9. Migrate From Multi Image Setup
10. Migrate Traefik from v2 to v3
11. PostgreSQL Upgrade
12. Troubleshooting Guides + NGINX in Windows + ARM64 / Apple Silicon

# Lessons and Tutorials
2. Play with Docker (Single Compose Setup)
3. Setup Examples
4. Multi Bench Example
5. Multi Tenant Setup

# Reference
1. Production Overview (Setup Overview)
2. Environment Variables
3. Compose Overrides
4. TLS/SSL Overview
5. Setting Up Debugger (Debugging)
6. Local Services
7. Configuring VitePress
8. Fork Management
9. Framework Comparisons
10. External Links
```

DanielRadlAMR:
Thanks for suggesting this! Here are my thoughts:

[...]

PR #1892, opened 2026-04-24
fix(build): invalidate cache when apps.json changes

### LINKED ISSUE #1887: Document cache behavior when using apps.json via BuildKit --secret
## Is your feature request related to a problem? Please describe.

When building custom images using `frappe_docker`, `apps.json` is passed via Docker BuildKit `--secret` to avoid leaking credentials (which is good).

However, in practice, changing `apps.json` (e.g. updating app versions, branches, or repositories) **does not invalidate Docker cache**, resulting in builds that:

* complete very quickly (~1–2 seconds)
* reuse previously cached layers
* **do not reflect the updated apps configuration**

This behavior is not documented and can lead to **silent inconsistencies between expected and actual build output**.

---

## Describe the solution you'd like

Add clear documentation explaining:

### 1. How `apps.json` is used

* It is mounted via BuildKit `--secret`
* It is not part of the Docker layer checksum

### 2. Cache implications

* Changes in `apps.json` do **not** invalidate cache
* Docker will reuse layers even if app versions or branches change

### 3. When this becomes a problem

* Updating app versions but build still uses old dependencies
* Debugging builds where output does not match configuration

### 4. Recommended workarounds

Include practical guidance such as:

* Use `--no-cache` when changing `apps.json`
* Use a cache-busting argument:

  ```bash
  docker build --build-arg CACHE_BUST=$(date +%s) ...
  ```
* (Optional) Use checksum-based invalidation:

  ```dockerfile
  ARG APPS_JSON_HASH
  RUN echo "$APPS_JSON_HASH"
  ```

---

## Describe alternatives you've considered

Alternative approaches considered:

* Using `COPY apps.json` instead of `--secret`

  * Pros: correct cache invalidation
  * Cons: risk of exposing credentials in image layers

* Keeping `--secret` but introducing explicit cache-busting mechanisms

  * More flexible, but requires documentation

---

## Additional context

This behavior is currently only observable by inspecting the Dockerfiles:

* `images/custom/Containerfile`
* `images/layered/Containerfile`

and noticing that:

```dockerfile
RUN --mount=type=secret,id=apps_json ...
```

Since BuildKit secrets are excluded from cache keys, Docker treats this layer as unchanged even when `apps.json` content differs.

---

## Why this matters

This is not just a performance detail — it can lead to:

* incorrect builds
* outdated dependencies
* confusion during deployment/debugging

Documenting this behavior would significantly improve developer experience and prevent subtle bugs.

---

jslocomotor:
## Summary

This updates the custom image build process to properly invalidate the cached bench init layer when apps.json changes while continuing to pass apps.json securely via BuildKit secrets.

Previously, changes to apps.json could be ignored by Docker layer caching because secret mounts do not participate in cache keys. As a result, custom app changes were not always reflected unless users would rebuilt with --no-cache.

This change introduces an optional APPS_JSON_HASH build argument in the affected Containerfiles. The hash is referenced during the bench init build step so cache invalidation happens automatically whenever apps.json changes, without exposing the file contents in image layers.

It also updates the build documentation with:

- manual build examples for Docker and Podman
- simple CI/CD pipeline example
- explanation of APPS_JSON_HASH
- clarification that apps.json must continue to be passed via --secret

## Why this approach
- keeps apps.json secret-safe
- avoids full rebuilds with --no-cache
- works for both manual and CI/CD builds
- preserves normal layer caching when apps.json is unchanged

## Testing

1.  Ran pre-commit run --all-files
2. Built images/layered/Containerfile successfully from scratch
3.  Rebuilt immediately without changing apps.json; build reused cache
4. Modified apps.json; the cached bench init layer was rebuilt, confirming cache invalidation
5. Applied same changes to images/custom/Containerfile
6. Validated documentation examples

Related Issues

Closes #1887

### DIFF
--- a/docs/02-setup/02-build-setup.md
+++ b/docs/02-setup/02-build-setup.md
@@ -7,7 +7,7 @@ This guide walks you through building Frappe images from the repository resource
 # Prerequisites
 
 - git
-- docker (Engine **v23.0+**) or podman
+- docker (Engine **v23.0+** with buildx) or podman
 - docker compose v2 or podman compose
 
 > Install containerization software according to the official maintainer documentation. Avoid package managers when not recommended, as they frequently cause compatibility issues.
@@ -23,7 +23,7 @@ cd frappe_docker
 
 # Define custom apps
 
-If you dont want to install specific apps to the image skip this section.
+If you don't want to include custom apps in the image, skip this section.
 
 To include custom apps in your image, create an `apps.json` file in the repository root:
 
@@ -31,11 +31,11 @@ To include custom apps in your image, create an `apps.json` file in the reposito
 [
   {
     "url": "https://github.com/frappe/erpnext",
-    "branch": "version-15"
+    "branch": "version-16"
   },
   {
     "url": "https://github.com/frappe/hrms",
-    "branch": "version-15"
+    "branch": "version-16"
   },
   {
     "url": "https://github.com/frappe/helpdesk",
@@ -44,7 +44,9 @@ To include custom apps in your image, create an `apps.json` file in the reposito
 ]
 ```
 
-# Build the image
+# Build custom images
+
+## Manually
 
 Choose the appropriate build command based on your container runtime and desired image type. This example builds the `layered` image with the custom `apps.json` you created.
 
@@ -55,9 +57,10 @@ Choose the appropriate build command based on your container runtime and desired
 ```bash
 docker build \
  --build-arg=FRAPPE_PATH=https://github.com/frappe/frappe \
- --build-arg=FRAPPE_BRANCH=version-15 \
+ --build-arg=FRAPPE_BRANCH=version-16 \
+ --build-arg=APPS_JSON_HASH="$(sha256sum apps.json | awk '{print $1}')" \
  --secret=id=apps_json,src=apps.json \
- --tag=custom:15 \
+ --tag=custom:16 \
  --file=images/layered/Containerfile .
 ```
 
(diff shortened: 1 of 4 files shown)


DanielRadlAMR:
The modified docs should stay purely manual.
Details regarding deployment with pipelines should have their own dedicated page, rather than being mixed into existing tutorials.

I think the “normal” documentation should be updated to recommend using `--no-cache`. This would make it much clearer what is actually happening. It might also make sense to add an additional reference section and link to it from the relevant point in the tutorial.

The `CACHE_BUST` approach really feels like something you would only want in pipeline-based workflows, so it should be documented only in the pipeline-specific documentation.

Furthermore, I would reconsider whether the hashed `apps.json` is the best choice for cache busting, or whether something more arbitrary like a timestamp or pipeline ID would be better. Even if `apps.json` does not change, you may still often want to pull updated apps from GitHub (for example, when a branch has been updated).

jslocomotor:
Sorry, following up your comment in #1887 I did lay out every possible cache busting procedure and this was the most elegant and self explanatory, without needing to modify CI/CD pipelines and while using the same Containerfile for both building methods as well.

I didn't mean to ignore your instructions, just tried different things while working on this and thought this would be the best way.

I am planning on a full documentation on CI/CD, for the time being I was more focused on fixing the behavior while still including the first hint in the current documentation. 
Waiting on #1843

Regarding pulling updates for apps if just specifying the branch you are right! Didn't catch it, because of pinning to releases for my build. 

I'll rework it to something like 
```yaml
ARG CACHE_BUST=""

RUN --mount=type=secret,id=apps_json,target=/opt/frappe/apps.json,uid=1000,gid=1000 \
  : "${CACHE_BUST}" && \
  ```
  
  So that everyone can implement his own technique: Hash, RunID, Date, etc.
  Manual builds will use `--no-cache` with the same Containerfile.

Rocket-Quack:
LGTM from a Containerfile side

One small docs/request before merge please clarify the `CACHE_BUST` behavior a bit more explicitly maybe you can create an extra page in the docs or give more examples and explanation for the usage

jslocomotor:
Thanks for checking! How about this approach?

I would kindly suggest quicker merging over ideal documentation because in the current release the repo is not fully CI/CD safe.

Rocket-Quack:
👍Merging

PR #1893, opened 2026-04-24
refactor(assets): externalize assets from sites volume using symlink

### LINKED ISSUE #1883: Asset mismatch between backend and frontend after running `bench migrate`

DanielRadlAMR:
Fixes #1883

https://github.com/frappe/frappe_docker/issues/1883#issuecomment-4316520500

This PR updates how assets are handled in our images. For details, refer to the linked issue and comment.


DanielRadlAMR:
This is an alternative to #1888

Only one of these PRs should be merged. I prefer this implementation.

images/custom/Containerfile
```diff
@@ -143,6 +143,11 @@ COPY --from=builder --chown=frappe:frappe /home/frappe/frappe-bench /home/frappe
 
 WORKDIR /home/frappe/frappe-bench
 
+# Move assets out of the sites volume, replace with symlink
+RUN cp -r /home/frappe/frappe-bench/sites/assets /home/frappe/frappe-bench/assets && \
+    rm -rf /home/frappe/frappe-bench/sites/assets && \
+    ln -s /home/frappe/frappe-bench/assets /home/frappe/frappe-bench/sites/assets
```
dandax123, line 149: This, and similar ones in the other dockerfiles won't work.   In Docker, if a path (like /home/frappe/frappe-bench/sites) is declared as a VOLUME, any changes made to that directory during the build process (like creating a symlink) are discarded once the container starts if they are overwritten by a mount.

dandax123: I created this #1894 which fixes this issue

DanielRadlAMR: This works as expected—I’ve verified it.

On fresh setups, it works out of the box. Existing setups require a one-time manual migration as described in the migration guide.

dandax123: Yeah, but we can automate the manual setup, which is what my PR does

dandax123: Also, I think this would only work on fresh setups, I don't think we should be expecting the user to manually migrate this. It makes sense that the image always have the setup done well. That's why I do the linking at runtime with an entry point

DanielRadlAMR: The setup itself doesn’t need to be manual—only the migration step, and only for existing installations.

I agree that manual work should be minimized. That said, some manual effort during major version upgrades is fairly common (for example with PostgreSQL).

Your approach avoids the need for a migration, but it comes at the cost of added complexity. In the long run, I think we should aim to reduce complexity where possible—it’s ultimately a trade-off.

Do you see it the same way, or do you have a different perspective?

---
I also understand that this situation can be frustrating. Right now, it feels like we’re introducing a new “major release” very frequently, which isn’t ideal.

My impression is that this is partly due to increased activity and attention on critical parts of the project, which is a good thing overall. However, our versioning may be a bit misleading—these releases are presented as stable, while in reality they behave more like pre-release iterations.

To be completely honest, many projects would likely still consider this stage as something closer to a v0.x phase.

DanielRadlAMR:
#1894 is yet another valid approach. Only one of both PRs should get merged.

Rocket-Quack:
This one is my favorite so i will merging this and close #1894

If necessary we can always switch later but I think this is the fastest and for now the most technically solid solution

dandax123:
> This one is my favorite so i will merging this and close #1894
> 
> If necessary we can always switch later but I think this is the fastest and for now the most technically solid solution

NO, please don't do that. they are not the same solution styles, please read the comments in #1894. This would break production for lots of people (like it did for me).  This would also manual migration basically on every bench updates, and / or modifications to compose.yml and helm charts.

Rocket-Quack:
> > This one is my favorite so i will merging this and close #1894
> > If necessary we can always switch later but I think this is the fastest and for now the most technically solid solution
> 
> NO, please don't do that. they are not the same solution styles, please read the comments in #1894. This would break production for lots of people (like it did for me). This would also manual migration basically on every bench updates, and / or modifications to compose.yml and helm charts.

I read the discussion and I agree that avoiding manual migration is a valid concern.
However I dont think it requires migration on every update. 
It should be a one-time migration for existing sites volumes "Only" replace the persisted `sites/assets` directory with a symlink
After that future image updates should use the image-owned assets automatically.
My point is not that #1893 requires no migration at all.

My main concern with #1894 is that it mixes the asset migration with unrelated Gunicorn configuration and adds runtime relinking on every container start. If automatic migration is preferred, I’d rather see that as a focused asset-only change.

If automatic migration is considered necessary for production safety I would prefer extracting that into a focused asset-only PR instead of merging it together with the Gunicorn changes.

From my POV: This is initially a merge to restore the ability to update images with migration steps. As I said, I’m not completely opposed to automated solutions they just shouldnt add an extra layer of complexity in this kind of speed. And as @DanielRadlAMR  already mentioned the Gunicorn changes should be included in a separate pull request because, in my opinion, they don’t currently address the site and asset mismatch issue.

I understand that it’s frustrating and since there are two solutions to the problem, I’m going with the one that seems simplest to me in the way of fixing fast the problem without complexity. 

I want to make it clear once again that your solution isn’t wrong it’s just that at this point, I find the changes required with Gunicorn too complicated for what is essentially an asset issue, and the automatic update, in my opinion, adds another layer of complexity.

docs/06-migration/01-assets-volume-change.md
```diff
@@ -0,0 +1,40 @@
+---
+title: Assets Volume Change
+---
+
+# Migration Guide: Assets Volume Change
+
+## Background
+
+The way `sites/assets` is handled has changed. Previously, assets were stored inside the a volume and persisted across container recreations. This could caused stale or mismatched assets after image updates. See [Assets Reference](../08-reference/07-how-assets-are-handeled.md) for details on the new approach.
```
Rocket-Quack, line 9: - Typo: "...the a volume..." remove "the"
- Reference Link `../08-reference/07-how-assets-are-handled.md` instead of `../08-reference/07-how-assets-are-handeled.md`

Rocket-Quack (changes requested):

dandax123:
> > > This one is my favorite so i will merging this and close #1894
> > > If necessary we can always switch later but I think this is the fastest and for now the most technically solid solution
> > 
> > 
> > NO, please don't do that. they are not the same solution styles, please read the comments in #1894. This would break production for lots of people (like it did for me). This would also manual migration basically on every bench updates, and / or modifications to compose.yml and helm charts.
> 
> I read the discussion and I agree that avoiding manual migration is a valid concern. However I dont think it requires migration on every update. It should be a one-time migration for existing sites volumes "Only" replace the persisted `sites/assets` directory with a symlink After that future image updates should use the image-owned assets automatically. My point is not that #1893 requires no migration at all.
> 
> My main concern with #1894 is that it mixes the asset migration with unrelated Gunicorn configuration and adds runtime relinking on every container start. If automatic migration is preferred, I’d rather see that as a focused asset-only change.
> 
> If automatic migration is considered necessary for production safety I would prefer extracting that into a focused asset-only PR instead of merging it together with the Gunicorn changes.
> 
> From my POV: This is initially a merge to restore the ability to update images with migration steps. As I said, I’m not completely opposed to automated solutions they just shouldnt add an extra layer of complexity in this kind of speed. And as @DanielRadlAMR already mentioned the Gunicorn changes should be included in a separate pull request because, in my opinion, they don’t currently address the site and asset mismatch issue.
> 
> I understand that it’s frustrating and since there are two solutions to the problem, I’m going with the one that seems simplest to me in the way of fixing fast the problem without complexity.
> 
> I want to make it clear once again that your solution isn’t wrong it’s just that at this point, I find the changes required with Gunicorn too complicated for what is essentially an asset issue, and the automatic update, in my opinion, adds another layer of complexity.

Okay, IMO, this is just way too much for a really simple problem. If you want another PR for gunicorn changes, I can do that, even though I don't think it's very necessary - I basically just took hardcoded values and made them configurable by ENV variables.  Again, this images are / can be used in various context, in my setup we use Kubernetes, and manual migration is NOT recommended, because the user of the image expects the assets to be there, even if I have to do that migration once, which isn't the case. And like I pointed in my first review comment that this DOESN'T fix the stale assets issue, except you do the migration every time you update the image assets, which is completely unnecessary to do manually. 

If you go with this solution, it means that we have to make PR to the helm charts (in the case of k8s), and other devops styles to accomodate this changes. 

>This, and similar ones in the other dockerfiles won't work. In Docker, if a path (like /home/frappe/frappe-bench/sites) is >declared as a VOLUME, any changes made to that directory during the build process (like creating a symlink) are discarded >once the container starts if they are overwritten by a mount.

Rocket-Quack:
I agree the K8s concern is valid. But the primary `frappe_docker` workflow is Docker/Compose and this gives us the cleanest asset model for that path (Despite the manual migration work I think this is acceptable for most users which are running Frappe on Docker) for now.

@dandax123 I dont consider that "fully" solved by this PR it should be coming a follow up with a migration change or a focused asset-only automatic migration PR for K8s setups that is also production safe for the default PVC setup from frappe/helm in case of runtime mutations. 

So in overview:
- fresh Docker/Compose setups can work with this change as-is
- existing volumes need migration
- K8s setups need an follow up PR (maybe an init container in frappe/helm or auto migration skripts

I would prefer not to merge the Gunicorn changes together with the asset fix. But the automatic migration concern from #1894 is valid and should be handled separately. 

If you still want to pursue the Gunicorn configurability please open that as a separate PR. If you want to pursue automatic asset migration, I’d also prefer that as a focused asset-only PR 🙂
Otherwise mixing them makes review, rollback, and release notes harder.

dandax123:
In summary of this, I've created both PRs: 

a. https://github.com/frappe/frappe_docker/pull/1896 -> for gunicorn changes
b. https://github.com/frappe/frappe_docker/pull/1894 ->  for fixing the CI

Again, k8s is not just where we have to do migration changes.  images are contracts, and with the current state the contract is broken, the users of this image expect the assets to be there (just like it was before). I can use dokploy, manual deployments, docker compose, k8s, etc to deploy this images, I'm arguing that it is completely unnecessary to break this contract and then give DevOps more manual work to accomodate a change, that is so trivial to fix. 

If I was managing 100 existing benches in a k8s setup, I would have to make yet another script to do manual migration after bench image updates

Rocket-Quack:
I understand your point and of course managing so many sites does add extra work in case of this manual migration. 
I will take another look at the PR and write you some improvments

Rocket-Quack:
After a productive discussion with @dandax123 and a further technical review #1894 was merged instead of this one.

In the event of a significant framework change the script can be removed again if necessary. 
Follow-up requests for improvements and to avoid errors and race conditions are also welcome should they arise.

PR #1894, opened 2026-04-25
fix(assets): auto-repair sites assets link on startup

dandax123:

This fixes the issue with the assets mismatch, and also includes a way to have variable gunicorn threads

### DIFF
--- a/images/custom/Containerfile
+++ b/images/custom/Containerfile
@@ -146,10 +146,22 @@ COPY --from=builder --chown=frappe:frappe /home/frappe/frappe-bench /home/frappe
 
 WORKDIR /home/frappe/frappe-bench
 
+# Move assets to image-layer storage
+RUN cp -r /home/frappe/frappe-bench/sites/assets /home/frappe/frappe-bench/assets && \
+  rm -rf /home/frappe/frappe-bench/sites/assets
+
 VOLUME [ \
   "/home/frappe/frappe-bench/sites", \
   "/home/frappe/frappe-bench/logs" \
 ]
+
+USER root
+# This entrypoint script link build assets of the image to the mounted sites volume at container initialization
+COPY resources/core/main-entrypoint.sh /usr/local/bin/entrypoint.sh
+RUN chmod +x /usr/local/bin/entrypoint.sh
+
+USER frappe
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
 
 CMD [ \
   "/home/frappe/frappe-bench/env/bin/gunicorn", \
--- a/images/layered/Containerfile
+++ b/images/layered/Containerfile
@@ -35,10 +35,22 @@ COPY --from=builder --chown=frappe:frappe /home/frappe/frappe-bench /home/frappe
 
 WORKDIR /home/frappe/frappe-bench
 
+# Move assets to image-layer storage
+RUN cp -r /home/frappe/frappe-bench/sites/assets /home/frappe/frappe-bench/assets && \
+  rm -rf /home/frappe/frappe-bench/sites/assets
+
 VOLUME [ \
   "/home/frappe/frappe-bench/sites", \
   "/home/frappe/frappe-bench/logs" \
 ]
+
+USER root
+# This entrypoint script link build assets of the image to the mounted sites volume at container initialization
+COPY resources/core/main-entrypoint.sh /usr/local/bin/entrypoint.sh
+RUN chmod +x /usr/local/bin/entrypoint.sh
+
+USER frappe
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
 
 CMD [ \
   "/home/frappe/frappe-bench/env/bin/gunicorn", \
--- a/images/production/Containerfile
+++ b/images/production/Containerfile
@@ -136,10 +136,22 @@ COPY --from=builder --chown=frappe:frappe /home/frappe/frappe-bench /home/frappe
 
 WORKDIR /home/frappe/frappe-bench
 
+# Move assets to image-layer storage
+RUN cp -r /home/frappe/frappe-bench/sites/assets /home/frappe/frappe-bench/assets && \
+  rm -rf /home/frappe/frappe-bench/sites/assets
+
 VOLUME [ \
   "/home/frappe/frappe-bench/sites", \
   "/home/frappe/frappe-bench/logs" \
 ]
+
+USER root
+# This entrypoint script link build assets of the image to the mounted sites volume at container initialization
+COPY resources/core/main-entrypoint.sh /usr/local/bin/entrypoint.sh
+RUN chmod +x /usr/local/bin/entrypoint.sh
+
+USER frappe
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
 
 CMD [ \
   "/home/frappe/frappe-bench/env/bin/gunicorn", \
--- a/resources/core/main-entrypoint.sh
+++ b/resources/core/main-entrypoint.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+set -e
+
+ASSETS_PATH="/home/frappe/frappe-bench/sites/assets"
+BAKED_PATH="/home/frappe/frappe-bench/assets"
+
+echo "Linking fresh assets to volume..."
+rm -rf "$ASSETS_PATH"
+mkdir -p "$(dirname "$ASSETS_PATH")"
+ln -s "$BAKED_PATH" "$ASSETS_PATH"
+
+exec "$@"


DanielRadlAMR:
This is essentially doing what #1888 implemented, but at build time instead of during compose up.

While you’re right that this approach also works—and wouldn’t require a migration—it makes the behavior less transparent and introduces additional complexity to the setup that doesn’t seem necessary. See: https://github.com/frappe/frappe_docker/issues/1883#issuecomment-4306502514

Regarding the Gunicorn changes: I’m not clear on what those modifications are doing. Could you open a separate PR containing only the Gunicorn-related changes, along with a more detailed explanation of their purpose?

dandax123:
> This is essentially doing what #1888 implemented, but at build time instead of during compose up.
> 
> While you’re right that this approach also works—and wouldn’t require a migration—it makes the behavior less transparent and introduces additional complexity to the setup that doesn’t seem necessary. See: [#1883 (comment)](https://github.com/frappe/frappe_docker/issues/1883#issuecomment-4306502514)
> 
> Regarding the Gunicorn changes: I’m not clear on what those modifications are doing. Could you open a separate PR containing only the Gunicorn-related changes, along with a more detailed explanation of their purpose?

No, it's not the same thing as #1888. We use this images in a kubernetes setup, which would require more unnecessary changes to support the #1888 approach. Also, the symlink doesn't happen at build time, the entrypoint script is executed first when a container starts, so this happens at run time actually.  It doesn't make the behavior less transparent, and actually doing the manual migration (or modifying compose, or helm charts) is a LOT more complex than this approach.

dandax123:
> Regarding the Gunicorn changes: I’m not clear on what those modifications are doing. Could you open a separate PR containing only the Gunicorn-related changes, along with a more detailed explanation of their purpose?

I modified the description of this PR to document more about the gunicorn changes

DanielRadlAMR:
> No, it's not the same thing as https://github.com/frappe/frappe_docker/pull/1888. We use this images in a kubernetes setup, which would require more unnecessary changes to support the https://github.com/frappe/frappe_docker/pull/1888 approach. Also, the symlink doesn't happen at build time, the entrypoint script is executed first when a container starts, so this happens at run time actually. It doesn't make the behavior less transparent, and actually doing the manual migration (or modifying compose, or helm charts) is a LOT more complex than this approach.

You’re right. The only thing I don’t completely like is that this introduces a check on every container start, which is a bit overkill for something only needed once per volume lifecycle. But the cost is minimal.

Please add small comments to the Containerfile.

Furthermore, since this makes our setup diverge from the standard Frappe architecture, please add reference documentation.
(For example: https://github.com/AMR-Mannesmann/frappe_docker/blob/refactor(assets)/externalize-assets-from-sites-volume-using-symlink/docs/08-reference/07-how-assets-are-handled.md)

DanielRadlAMR:
add GUNICORN build args here https://github.com/frappe/frappe_docker/blob/main/docs/02-setup/02-build-setup.md#build-args

dandax123:
> Please add small comments to the Containerfile.
I added comments

> Furthermore, since this makes our setup diverge from the standard Frappe architecture, please add reference documentation. (For example: https://github.com/AMR-Mannesmann/frappe_docker/blob/refactor(assets)/externalize-assets-from-sites-volume-using-symlink/docs/08-reference/07-how-assets-are-handled.md)

No, this doesn't actually diverge from a standard frappe architecture, the `assets` in the `sites\assets` is the standard architecture for how a bench looks

dandax123:
> add GUNICORN build args here https://github.com/frappe/frappe_docker/blob/main/docs/02-setup/02-build-setup.md#build-args

Using a build arg here only makes configuration easier, and not the solution that we want, because that would require building another image to be able to change the number of workers. The idea with the env variables is that you can restart your container with a new value for a `GUNICORN_X` variable on the same image 

Basically, we want to be able to configure gunicorn stuff at run time, not at build time

Rocket-Quack:
Would it make sense to make the startup migration a bit more defensive before merging? 
Right now the entrypoint skript removes and recreates `sites/assets` on every container start. 
Since that path is inside the mounted `sites` volume i would prefer to avoid mutation of persistent runtime data.

This isn't about the approach itself but since we are using `rm-rf` we should handle it as defensively as possible.
I agree the risk is limited but Still it lives inside the mounted `sites` volume.

TODOs @dandax123 
Please move the entrypoint skript into a dedicated script file instead of embedding it as a heredoc in every containerfile. 
This make the migration easier to review, test, validate with ShellCheck and reuse across images.

Could you also add a documentation? It should explain what exactly happens to `sites/assets` and what existing deployments should expect after updating.

dandax123:
> Would it make sense to make the startup migration a bit more defensive before merging? Right now the entrypoint skript removes and recreates `sites/assets` on every container start. Since that path is inside the mounted `sites` volume i would prefer to avoid mutation of persistent runtime data.

Not necessarily, the idea is that the assets are built-time artifacts. These assets would always be available in the image, and this is the only way to ensure that the assets served are not stale.

> TODOs @dandax123 Please move the entrypoint skript into a dedicated script file instead of embedding it as a heredoc in every containerfile. This make the migration easier to review, test, validate with ShellCheck and reuse across images.

Done 

> Could you also add a documentation? It should explain what exactly happens to `sites/assets` and what existing deployments should expect after updating.

Not really, there is no behavioral change actually. So for both new and current deployments, it'll work exactly as expected.

Rocket-Quack:
Thank you for the productive and objective discussion. 

At first, I was more in favor of merging the other PR but after reviewing this PR I have come to the conclusion that we can now take this approach especially since, like you argued, the automated method is a better fit here. 

However we should continue to address the underlying issue on the Frappe Framework side so that we can safely remove this extra layer in the future.

At the same time we should keep race safety in mind, especially around startup and submit further PRs if necessary.

@dandax123 please fix lint workflow issues then its ready for merging 👍

dandax123:
> @dandax123 please fix lint workflow issues then its ready for merging 👍

Done

Rocket-Quack:
All right thanks. 
@dandax123 you were able to test this in a Compose setup with multiple containers starting up that use the same image so the likelihood of a race condition occurring is firstly low.

Merging this then in favor of #1893 and closing #1883 but still the option to enhance this script and harden it to dismiss unsave deletions.

jslocomotor:
Wait, wait, wait guys. How is it, that my PRs get hyper checked regarding documentation and here NOTHING is written?

This is BREAKING! My build repo i.e. just includes the Containerfile, not the whole frappe_docker repo. So no resources/core/main-entrypoint.sh available.

dandax123:
> Wait, wait, wait guys. How is it, that my PRs get hyper checked regarding documentation and here NOTHING is written?
> 
> This is BRAKING! My build repo i.e. just includes the Containerfile, not the whole frappe_docker repo. So no resources/core/main-entrypoint.sh available.

In the dockerfile there are already 3 COPY instructions from the resources directory,  so another COPY instruction can't break your repo.

jslocomotor:
Please tell that my CI/CD pipeline:

```
Containerfile:49
--------------------
  47 |     USER root
  48 |     # This entrypoint script link build assets of the image to the mounted sites volume at container initialization
  49 | >>> COPY resources/core/main-entrypoint.sh /usr/local/bin/entrypoint.sh
  50 |     RUN chmod +x /usr/local/bin/entrypoint.sh
  51 |     
--------------------
ERROR: failed to build: failed to solve: failed to compute cache key: failed to calculate checksum of ref kwgqhkbi9qbx564gonc58rze2::m6cvwqn66vt8zvjro4u19dq34: "/resources/core/main-entrypoint.sh": not found
⚙️ [runner]: exitcode '1': failure```

jslocomotor:
I'm using the layered Containerfile, that is why this is breaking for me. First `COPY` from resources in there.

dandax123:
> I'm using the layered Containerfile, that is why this is breaking for me. First `COPY` from resources in there.

I see, I'll write a fix for you

dandax123:
> > I'm using the layered Containerfile, that is why this is breaking for me. First `COPY` from resources in there.
> 
> I see, I'll write a fix for you

Done: https://github.com/frappe/frappe_docker/pull/1898

DanielRadlAMR:
> No, this doesn't actually diverge from a standard frappe architecture, the `assets` in the `sites\assets` is the standard architecture for how a bench looks

Yeah, but usually `sites/assets` is not a symlink and `assets` does not exists separately. It would be helpful to document this behavior explicitly.



> Using a build arg here only makes configuration easier, and not the solution that we want, because that would require building another image to be able to change the number of workers. The idea with the env variables is that you can restart your container with a new value for a `GUNICORN_X` variable on the same image

I meant that documentation for the new variables should be added. If I referenced the wrong place, please update the appropriate documentation accordingly.

DanielRadlAMR:
@jslocomotor 

> Wait, wait, wait guys. How is it, that my PRs get hyper checked regarding documentation and here NOTHING is written?

Sorry about your experience — that’s a fair point. We do try to consistently review PRs for missing documentation, regardless of who submits them.

That said, there have been quite a few changes recently, and since we’re all contributing alongside our day-to-day work, some things do slip through.I will try to improve our standards going forward.

Thanks for staying constructive, even though this caused issues for your setup. ❤️

jslocomotor:
> @jslocomotor
> 
> > Wait, wait, wait guys. How is it, that my PRs get hyper checked regarding documentation and here NOTHING is written?
> 
> Sorry about your experience — that’s a fair point. We do try to consistently review PRs for missing documentation, regardless of who submits them.
> 
> That said, there have been quite a few changes recently, and since we’re all contributing alongside our day-to-day work, some things do slip through.I will try to improve our standards going forward.
> 
> Thanks for staying constructive, even though this caused issues for your setup. ❤️

No hard feelings at all. Just came home to see my pipeline failing, then found this merge and raged a little bit :smile: 

You and @Rocket-Quack are doing a great job keeping this project aligned and from what I've seen the past weeks, your standards are already high, which is a good thing. I have nothing but nice words to say about both of you as maintainers. What really stands out is the time and effort you put into every issue, PR, code review and conversations in general. Please keep that up, as it motivates to be part of this project!

PR #1896, opened 2026-04-27
feat: configure gunicorn with env variables

dandax123:
This PR also adds supports for customizing gunicorn values. It is reasonable to be able to configure the number of gunicorn workers and thread that you have, and this PR supports that by adding environment variables to do so.

For instance, you can have a bench with 10+ sites, and 2 gunicorn workers is not enough, there's no reason to have to manually modify the hardcoded dockerfile to increment the workers, you should just be able to configure that by environment variables.

### DIFF
--- a/images/production/Containerfile
+++ b/images/production/Containerfile
@@ -156,18 +156,10 @@ USER root
 COPY resources/core/main-entrypoint.sh /usr/local/bin/entrypoint.sh
 RUN chmod 755 /usr/local/bin/entrypoint.sh
 
+COPY resources/core/start.sh /usr/local/bin/start.sh
+RUN chmod 755 /usr/local/bin/start.sh
+
 USER frappe
 ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
 
-CMD [ \
-  "/home/frappe/frappe-bench/env/bin/gunicorn", \
-  "--chdir=/home/frappe/frappe-bench/sites", \
-  "--bind=0.0.0.0:8000", \
-  "--threads=4", \
-  "--workers=2", \
-  "--worker-class=gthread", \
-  "--worker-tmp-dir=/dev/shm", \
-  "--timeout=120", \
-  "--preload", \
-  "frappe.app:application" \
-]
+CMD ["start.sh"]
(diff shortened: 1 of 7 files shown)


DanielRadlAMR:
Please fix the lint errors and add documentation for the new variables.

images/production/Containerfile
```diff
@@ -153,15 +158,13 @@ RUN chmod +x /usr/local/bin/entrypoint.sh
 USER frappe
 ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
 
-CMD [ \
```
DanielRadlAMR, line 162: Is there a reason why you did move away from JSON fromat?

DanielRadlAMR: Is this because the variables cannot be passed in JSON format?

I also think this is somewhat related to the discussion in #1898. I would suggest moving the CMD block into a separate file, similar to how the entrypoint is handled.

dandax123: No, it is not related to #1898.

>Is this because the variables cannot be passed in JSON format?

Yes, the new form is the shell mode, which allows variable substitutions that are necessary for what we want to achieve. In the previous mode (exec), that won't allow the substitution.  Also, we can't move a CMD block to a separate file. 

More details here: https://emmer.dev/blog/docker-shell-vs.-exec-form/

DanielRadlAMR: Well in that case it is related to #1898

Move it into a shell script, like the entrypoint

dandax123: addressed

DanielRadlAMR:
If you want to keep this with environment variables, a few things still need to be done:

* add documentation: https://github.com/frappe/frappe_docker/blob/main/docs/02-setup/04-env-variables.md
* pass the new variables to the `backend` service in the compose file so they can be overridden at runtime
* optional: add the variables to `example.env`

dandax123:
> If you want to keep this with environment variables, a few things still need to be done:
> 
>     * add documentation: https://github.com/frappe/frappe_docker/blob/main/docs/02-setup/04-env-variables.md
> 
>     * pass the new variables to the `backend` service in the compose file so they can be overridden at runtime
> 
>     * optional: add the variables to `example.env`

Done. Apologies for the delay, I was quite busy in the past few days

DanielRadlAMR:
`v15` build fails do have a different reason: https://github.com/frappe/frappe/pull/39253

PR #1897, opened 2026-04-29
feat(compose): add migrator service override and documentation

### LINKED ISSUE #1878: Auto migration after app update
We use an automated deployment for our ERPNext instances. When modifying custom images, i.e. updating apps, it is best practice to run `bench migrate` in order to update the database schema or DocTypes.
Most instructions show executing the command from docker host or backend container, which interferes with automated deployment.

**Describe the solution you'd like**
Find a good way to automate migration. Our current approach:
Adding a "migrator" container to the stack:
```yaml
migrator:
  <<: *backend_defaults
  entrypoint:
    - bash
    - -c

  command:
    - >
      if [ -d "sites/$$SITE_NAME" ]; then
        bench --site $$SITE_NAME migrate;
      else
        echo "Site $$SITE_NAME does not exist yet, skipping migrate.";
      fi

  depends_on:
    configurator:
      condition: service_completed_successfully

    db:
      condition: service_healthy

    redis-cache:
      condition: service_started

    redis-queue:
      condition: service_started

  restart: on-failure
```
This will run once after startup or update, check if the defined site exists and then run migration.

Feel free to point out other/ better solutions or problems with the above approach.

jslocomotor:
## Summary

This introduces an optional `compose.migrator.yaml` override that adds a dedicated migrator service.

The service runs `bench migrate` automatically on a given site during stack startup, making it easier to handle database and schema updates in automated deployments.

Additionally, the override is documented and integrated into the existing overrides and deployment documentation.

---

## Motivation

In automated build and deployment setups, running `bench migrate` manually inside the backend container is not practical.

Without running migrations after updating images or apps, the site may become inconsistent or fail to start properly.

This change provides a simple and optional solution to automate this step.

Closes #1878 

---

## Implementation

* Adds `overrides/compose.migrator.yaml`

  * runs `bench --site $SITE_NAME migrate` on startup
  * skips execution if the site does not exist
  * fails early if `SITE_NAME` is not set
  * uses `restart: on-failure` for robustness

* Uses the same image and configuration pattern as backend services

* Includes consistent logging for easier debugging

---

## Documentation

* Adds `compose.migrator.yaml` to the overrides overview
* Adds a dedicated section describing its behavior and usage
* References the migrator in the automated deployment guide

---

## Usage

```bash
docker compose \
  -f compose.yaml \
  -f overrides/compose.migrator.yaml \
  up -d
```

Requires:

```bash
SITE_NAME=your.site.name
```

---

## Behavior

* exits with error if `SITE_NAME` is not set
* skips migration if the site does not exist yet
* runs `bench migrate` if the site exists

---

## Testing

* validated with `docker compose config`
* tested startup behavior with:

  * missing `SITE_NAME`
  * non-existing site
  * existing site
* ran `pre-commit run --all-files`

---

## Notes

* The migrator service is optional and does not affect existing setups
* Designed primarily for CI/CD and automated deployment scenarios


overrides/compose.migrator.yaml
```diff
@@ -0,0 +1,45 @@
...
+        fi;
+        if [ -d "sites/$$SITE_NAME" ]; then
+          echo "[migrator] Migrating $$SITE_NAME";
+          bench --site $$SITE_NAME migrate;
+        else
+          echo "[migrator] Site $$SITE_NAME not found, skipping migrate";
+        fi
+    environment:
+      SITE_NAME: ${SITE_NAME}
+    restart: on-failure
```
Rocket-Quack: If you do not set `SITE_NAME` the container would terminate with code 1 (error) but would then be restarted resulting in an endless loop of the container starting and stopping. 

Please specify either the maximum number of errors allowed (example: on-failure:5) or remove it entirely using “no”
Another option would be to handle this directly via environment variables so that the missing `SITE_NAME` would be detected when running `docker compose up/config`

docs/03-production/06-automated-builds-and-deployment.md
```diff
@@ -126,3 +126,22 @@ Note: When using branch references in `apps.json`, the hash only changes when th
       --tag=custom:16 \
       --file=images/layered/Containerfile .
 ```
+
+## Automated deployment
+
+### Automate site migration
+
```
Rocket-Quack, line 133: Please note here that the .env file must be modified when using this, and specify what the input should be

Add this to env. `SITE_NAME=my.example.domain`
Perhaps by also give a command appending it to the end of the env file as is already done here: 
https://github.com/frappe/frappe_docker/blob/473f08a7f9ed672bd2a5782f9527c6fc6de00209/docs/02-setup/08-single-server-nginxproxy-example.md?plain=1#L117

Rocket-Quack (changes requested):

Rocket-Quack:
This currently only covers single-site setups because the migrator runs `bench migrate` for the one configured `SITE_NAME` in the .env

If that is what you wanted it should either be documented explicitly in the overrides overview and docs or the implementation should be extended to support multi-site stacks as well (i would prefer that)
 
Otherwise users maybe running multiple sites in the same stack assume that all sites are migrated automatically while only one site is actually handled.

How do you feel about this @jslocomotor?

jslocomotor:
Thanks for your feedback, really appreciate it!

As I don't like the idea of multi-tenancy, nor quite understand why someone would run something so lightweight not in separate stacks (allowing for different images, apps, backup strategies, separated restores, etc.), Frappe and this repo still enable this feature. And my service should be compatible.

One approach would be using `bench --site all migrate` which will of course work with one or multiple sites. And since the same image will be shared, a mass migration is mandatory, right?

Rocket-Quack:
Exactly
I would use `bench --site all migrate` 
In the environment then the value to set either true or false, and then you can enable or disable the Migrator service as needed.

docs/03-production/06-automated-builds-and-deployment.md
jslocomotor: `SITE_NAME` no longer needed.

overrides/compose.migrator.yaml
jslocomotor: Added 5 retry attempts before exiting. Implemented more robust error handling.

jslocomotor:
Changed command to use `bench --site all migrate` . Integrated lean debugging and error handling.
`MIGRATE_SITES` is used for switching, defaults to `true`.

Rocket-Quack:
Please also update the documentation for the newly introduced env variable at 
https://github.com/frappe/frappe_docker/blob/main/docs/02-setup/04-env-variables.md

That is the central env variable reference feel free to add a new section there such as "Migration Service"

jslocomotor:
Got me, missed that. Took a look at https://github.com/frappe/frappe_docker/blob/main/example.env as well but figured that only non defaulting variables from overrides need to be added.

PR #1898, opened 2026-05-01
fix: embed entrypoint for layered image

dandax123:
the `images\layered` uses an existing frappe build as base image. With that, it doesn't use the context of `frappe_docker` for the build context, this cause the local COPY instructions to fail. 

This PR replaces the copy of the entrypoint to be embedded into the Dockerfile.

### DIFF
--- a/images/layered/Containerfile
+++ b/images/layered/Containerfile
@@ -46,8 +46,21 @@ VOLUME [ \
 
 USER root
 # This entrypoint script link build assets of the image to the mounted sites volume at container initialization
-COPY resources/core/main-entrypoint.sh /usr/local/bin/entrypoint.sh
-RUN chmod +x /usr/local/bin/entrypoint.sh
+RUN cat <<'EOF' > /usr/local/bin/entrypoint.sh \
+ && chmod +x /usr/local/bin/entrypoint.sh
+#!/bin/bash
+set -e
+
+ASSETS_PATH="/home/frappe/frappe-bench/sites/assets"
+BAKED_PATH="/home/frappe/frappe-bench/assets"
+
+echo "Linking fresh assets to volume..."
+rm -rf "$ASSETS_PATH"
+mkdir -p "$(dirname "$ASSETS_PATH")"
+ln -s "$BAKED_PATH" "$ASSETS_PATH"
+
+exec "$@"
+EOF
 
 USER frappe
 ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]


jslocomotor:
Thank you, can confirm that this works out nicely. Even when just using the Containerfile alone, outside this repo.

Rocket-Quack:
I have to interject here quickly. 
On the one hand, I understand that you might only use the build context you need from the repo but as far as I know, the documentation doesnt assume standalone support for layered Dockerfiles anywhere. 

From my perspective, in addition to heredoc—which, as mentioned in the other PR, isn’t ideal for testing, etc.—this also introduces duplicates of the script already present in the core scripts folder. 

If someone just copies the Dockerfile, they’re tying themselves to a random current implementation detail. As soon as layered needs to COPY resources/foo, config templates, or a helper script tomorrow, this pipeline breaks again.

In my view the used building pipeline being used needs to be adapted and the build context should not just use a single Dockerfile but as described in the documentation the Docker repo along with the stored scripts.

I don't know about you @DanielRadlAMR but from my perspective this is something fundamental that wasn't intended to be used as a standalone Dockerfile. The build context is the Docker repository. 

So, from my point of view, I wouldn't merge it. 
Sorry

jslocomotor:
Thanks for adding your valid opinion.
This might be the right decision, maybe I'm the only one who does it that way. 
Just to explain why: 

1. Security. One could hide some harmful code inside the repo and inject it. - Not that I expect someone to do so. The Containerfile can be checked easily upon applying changes, the whole repo not.
2. When my build fails after updating, I only have to look at one file, not a context of files.
3. My build repo doubles as my deployment repo and I don't want all the unnecessary files ending up on my docker host. (Also for 1.)

This may be closed for keeping the repo clean then. Thanks @dandax123 for the discussion and quick solution finding.

DanielRadlAMR:
@jslocomotor

1. One could also add harmful code directly to the Containerfile, so it doesn’t really make much of a difference whether the entrypoint is included as a file or directly as code.
2. I’m not quite sure I understand this point.
3. I also don’t quite follow this — what files would end up on the Docker host? Typically you just push the image itself to the host. Since the image stays more or less the same whether you copy `frappe_docker` or use a Containerfile directly, I don’t see a practical difference here.

FYI — this might be useful for you:
I also have a customized build and deployment process that I manage in a separate repository. Since it depends heavily on `frappe_docker`, I’ve included it as a Git submodule in my deployment repo. This makes managing different versions in the deployment process quite convenient.

DanielRadlAMR:
I mostly agree with @Rocket-Quack.

This repo expects the entire repository to be cloned. If only parts are cherry-picked, this deviates from the documented/expected setup, so we cannot reliably account for those cases since we have no way of knowing what gets selected.

There may still be valid reasons to implement the containerfiles differently and avoid using an entrypoint script in favor of “pure code”. However, if we decide to change this behavior, it should be applied consistently across all Containerfiles, and the script should be removed entirely.

At the moment, I cannot think of a strong reason why the alternative approach would be preferable. In my opinion, the current approach is cleaner for two reasons:

* Better readability of the Containerfile
* A single entrypoint script instead of duplicating hardcoded entrypoints across multiple Containerfiles (better maintainability)

Rocket-Quack:
Closing this PR then

PR #1900, opened 2026-05-04
Add pycups as optional install feature

### LINKED ISSUE #1895: Optional CUPS utils on custom image
I am interested in adding "Silent printing"/"Backend printing" functionality to my Docker install. 
I stumbled upon the docs here: https://docs.frappe.io/erpnext/print-settings#3-network-printer-print-server
As a test, I took my running container and installed `libcups2-dev` and `cups` packages (still don't know if cups itself is needed). After running `./env/bin/pip install pycups` and then `bench restart`, looks like a success - Frappe connects to CUPS:
Image
(it's definitely connecting becauses the printer name was fetched from the CUPS service)

Obviously it's just a test setup since the packages will not be retained in the container this way. 

My question is: could we add the required packages and commands to dockerfile as optional, dependant on some build arg (e.g. INCLUDE_CUPS)? Of course it wouldn't be a good idea to add it for everyone since these packages easily add like 600MB of data. 

I am definitely not an expert in dockerfiles so can't offer much contribution  myself, but I'm happy to test things.

**Describe alternatives you've considered**
A different approach is to not use built-in function at all, instead to build separate module for generating PDFs and sending it to a dedicated container managing print jobs. A bit more convoluted. 

Keep in mind all of that requires further testing, I just wanted to stop at "I think it could work".

martkaczmarek:
This PR adds option to install pycups and dependencies to custom docker images, using INSTALL_PYCUPS build argument. 
Resolves #1895. 

Todo:
'layered' image changes
`RUN /home/frappe/frappe-bench/env/bin/pip install --no-cache-dir pycups` command

### DIFF
--- a/images/custom/Containerfile
+++ b/images/custom/Containerfile
@@ -123,7 +123,7 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 # Install pycups
-RUN if [ "$INSTALL_PYCUPS" != "false" ]; then RUN /home/frappe/frappe-bench/env/bin/pip install --no-cache-dir pycups; fi \
+RUN if [ "$INSTALL_PYCUPS" != "false" ]; then RUN /home/frappe/frappe-bench/env/bin/pip install --no-cache-dir pycups; fi
 
 USER frappe
 
--- a/images/custom/Containerfile
+++ b/images/custom/Containerfile
@@ -64,7 +64,7 @@ RUN useradd -ms /bin/bash frappe \
     && apt-get install -y ./$downloaded_file \
     && rm $downloaded_file \
     # Install cups 
-    && if [ "${INSTALL_PYCUPS}" != "false" ]; then apt-get install --no-install-recommends -y libcups2-dev cups-client fi \
+    && if [ "$INSTALL_PYCUPS" != "false" ]; then apt-get install --no-install-recommends -y libcups2-dev cups-client fi \
     # Chromium
     && if [ "$INSTALL_CHROMIUM" != "false" ]; then \
         DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
@@ -123,7 +123,7 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 # Install pycups
-RUN if [ "${INSTALL_PYCUPS}" != "false" ]; then RUN /home/frappe/frappe-bench/env/bin/pip install --no-cache-dir pycups fi \
+RUN if [ "$INSTALL_PYCUPS" != "false" ]; then RUN /home/frappe/frappe-bench/env/bin/pip install --no-cache-dir pycups fi \
 
 USER frappe
 
--- a/images/custom/Containerfile
+++ b/images/custom/Containerfile
@@ -76,7 +76,7 @@ RUN useradd -ms /bin/bash frappe \
     && mkdir -p /etc/nginx/snippets \
     && pip3 install frappe-bench \
     # Install pycups
-    && if [ "$INSTALL_PYCUPS" != "false" ]; then /home/frappe/frappe-bench/env/bin/pip install --no-cache-dir pycups; fi \
+    && if [ "$INSTALL_PYCUPS" != "false" ]; then pip install --no-cache-dir pycups; fi \
     # Fixes for non-root nginx and logs to stdout
     && sed -i '/user www-data/d' /etc/nginx/nginx.conf \
     && ln -sf /dev/stdout /var/log/nginx/access.log && ln -sf /dev/stderr /var/log/nginx/error.log \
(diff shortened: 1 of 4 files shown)


images/custom/Containerfile
```diff
@@ -62,6 +63,8 @@ RUN useradd -ms /bin/bash frappe \
     && curl -sLO https://github.com/wkhtmltopdf/packaging/releases/download/$WKHTMLTOPDF_VERSION/$downloaded_file \
     && apt-get install -y ./$downloaded_file \
     && rm $downloaded_file \
+    # Install cups 
+    && if [ "${INSTALL_PYCUPS}" != "false" ]; then apt-get install --no-install-recommends -y libcups2-dev cups-client fi \
```
DanielRadlAMR: i'd use `= "true"` here

DanielRadlAMR:
I think this should be added to the bench containerfile as well

DanielRadlAMR:
It might be helpful to add a “Print Server” setup example in `02-setup/06-setup-examples.md`.

DanielRadlAMR:
@martkaczmarek do you plan to update this PR?

martkaczmarek:
@DanielRadlAMR sorry about that; I still need to finish and test everything on my end. 
 I'll try to update it this weekend or next week.
Although I would still appreciate help with positioning of the RUN command listed in the furst comment - should it be done right at the end of current dockerfiles?

Also, about docs: are you sure `02-setup/06-setup-examples.md` is a good place for that? This document is about using different compose configs, not about building a container. `02-setup/02-build-setup.md` explains fairly well that for custom container componentss (Chromium, Python versions, Cups) you should build a container and  pass its name as CUSTOM_IMAGE=xyz in .env

DanielRadlAMR (changes requested):

DanielRadlAMR:
Hey again. 

> Although I would still appreciate help with positioning of the RUN command listed in the furst comment - should it be done right at the end of current dockerfiles?

I'm not exactly sure what you mean here.

---

Regarding the documentation, I think we should document somewhere how to enable the print server, since this is quite specific and difficult to figure out if it is not documented properly.

What I wrote was only a suggestion though, so feel free to share your idea on that.

Or do you think that adding:

`Set true to install pycups and its dependencies; see https://docs.frappe.io/erpnext/print-settings`

to the build args is already enough documentation for others to understand it?

images/custom/Containerfile
martkaczmarek: Sure, I'll change that in all 3 files next commit

martkaczmarek:
> > Although I would still appreciate help with positioning of the RUN command listed in the furst comment - should it be done right at the end of current dockerfiles?
> 
> I'm not exactly sure what you mean here.

This command:
`RUN /home/frappe/frappe-bench/env/bin/pip install --no-cache-dir pycups`
as per official Frappe documentation, needs to be run somewhere in the end of all install steps - PIP needs to be available, and i *think* frappe/bench needs to be already present as well. 

For now the only version I managed to put into dockerfile without build errors is this:
`&& if [ "$INSTALL_PYCUPS" != "false" ]; then pip install --no-cache-dir pycups; fi \`
BUT: it's either run with a wrong user, in wrong directory, wrong Python environment... something like that. The build finishes, but after starting the container, Frappe doesn't see pycups. If I *then* go into the container as root and run `/home/frappe/frappe-bench/env/bin/pip install --no-cache-dir pycups`, then everything starts working (the underlying binary libraries installed with apt are fine). 

I don't know enough about Python structure to figure out when in the build process (or with what user?) the `pip install pycups` should be run. 

About the docs: I really think the build args manual is the only suitable place. It's exactly the same as headless Chrome installation argument - it only matters in the build stage. After installing pycups and libraries inside the dockerfile, configuration is done and user should follow Frappe docs to configure printers in the app itself. 
If that's possible, after implementing it here, we could propose a mention of Docker build argument in the main Frappe docs, just like they point to required libraries for non-docker installs. 

Also: I do realize commits in this PR are a mess, I really should have tested it in a separate branch. After finishing here, I will open a separate PR with condensed commits.

DanielRadlAMR:
Thanks for the further explanation.

TBH, I cannot tell you anything for sure here since I never used this myself — but I guess the best thing to do is some trial and error.

Nevertheless, I just checked the docs, and I think you already found the solution.

Currently:

```yaml
# Install pycups
&& if [ "$INSTALL_PYCUPS" != "false" ]; then pip install --no-cache-dir pycups; fi \
```

But I think you need something like this instead:

```dockerfile
RUN /home/frappe/frappe-bench/env/bin/pip install --no-cache-dir pycups
```

(as you already mentioned in your previous comment)

The difference here is *which* `pip` is being used. Currently, you are not inside the bench folder, so `pycups` gets installed into the system Python interpreter instead of the bench virtual environment.

You either need to:

* navigate into the `frappe-bench` directory before running `pip install ...`, or
* explicitly specify which `pip` binary should be used (like you did in the `RUN` command by providing the full path).

---

Okay, I guess no further docs are needed then. Just do me a favor and add a small note to the build args that `pycups` is required for the print server.

For me, this was not obvious without checking other docs, so a short note there would make the build arg much clearer :)

martkaczmarek:
I'm closing this for now as I didn't have time to work on the feature. I'll reopen as new PR when I have something to show.

PR #1902, opened 2026-05-05
docs: add CLAUDE.md with codebase guidance for Claude Code

MarlonVelezB:
## Summary

- Adds `CLAUDE.md` to the repository root with guidance for Claude Code when working in this codebase
- Documents common commands (dev servers, asset builds, tests, linting, site management)
- Describes high-level architecture: multi-app structure, DocType system, hooks extensibility, asset pipeline, background jobs, and multi-tenancy

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/development/frappe-bench/CLAUDE.md
+++ b/development/frappe-bench/CLAUDE.md
@@ -1,0 +1,123 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## What This Repo Is
+
+This is a **frappe-bench** — a containerized development workspace that manages multiple Frappe Framework applications. Frappe is a full-stack Python web framework (Flask/Werkzeug-based) with a built-in DocType ORM, Vue.js 3 frontend, Redis-backed queues/cache, and real-time Socket.io communication.
+
+**Installed apps** (under `apps/`):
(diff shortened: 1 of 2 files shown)


DanielRadlAMR:
I’m not entirely convinced this is the direction we want to take. I agree with the intention—improving the AI development experience for Frappe is definitely valuable—but I don’t have enough experience with these workflows to judge whether this is the right approach.

There are already external efforts, such as https://github.com/OpenAEC-Foundation/Frappe_Claude_Skill_Package, which focus specifically on managing skills for Frappe development. Developers who want this kind of setup can already integrate it into their own environments.

More broadly, this also touches on separation of concerns: this repository should primarily focus on deployment, and everything related to deployment should live here. Enhancing the AI development experience doesn’t seem directly related to that scope and might be better handled in a dedicated project.

That said, it could be a good idea to expand the development documentation to include guidance on using AI tools alongside our setup.

Rocket-Quack:
I'm not sure if this was even intended as a PR. The fact that a development setup was committed—including apps and an AI Markdown file in the typical Docker Frappe development directory setup—and the added app are things I'm personally unfamiliar with and shouldn't be included in a general dev setup. I also agree with @DanielRadlAMR and don't see the point in adding skills for AIs here.

PR #1903, opened 2026-05-06
Fix entrypoint.sh permission

oktett-8:
Testing with FRAPPE_VERSION=v16.17.4 start of containers failed with error in docker logs:
`/bin/bash: /usr/local/bin/entrypoint.sh: Permission denied`

Cause was a permission setting for this file of `-rwx--x--x`

To avoid execution permission problems (even with umask set) `chmod +x` was changed to `chmod ugo+rx` in `images/layered/Containerfile` resulting in `-rwxr-xr-x` for `/usr/local/bin/entrypoint.sh` inside the relevant containers, so they can start without problem.

### DIFF
--- a/docs/02-setup/02-build-setup.md
+++ b/docs/02-setup/02-build-setup.md
@@ -96,6 +96,7 @@ See [Automated Builds and Deployment](../03-production/06-automated-builds-and-d
 | PYTHON_VERSION       | Python version for the base image                                                                                                 |
 | NODE_VERSION         | Node.js version                                                                                                                   |
 | WKHTMLTOPDF_VERSION  | wkhtmltopdf version                                                                                                               |
+| INSTALL_CHROMIUM     | Configure chromium installation, defaults to `true` - needed for Frappe Workbench version >15                                     |
 | **bench only**       |                                                                                                                                   |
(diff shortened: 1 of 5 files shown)


Rocket-Quack:
Please update your branch onto current main
Right now the PR appears to revert newer changes 😀

oktett-8:
Yes, sorry, got interrupted yesterday and messed up while trying to re-write history ;-)

I just reverted the reverts. It’s now the “755”-version applied to all nginx-entrypoint.sh and entrypoint.sh I found.

varun-krishnamurthy:
@oktett-8  Can you update the code to remove the files under the assets instead of the whole asset folder? Our Frappe Operator https://github.com/vyogotech/frappe-operator does the same thing but doesn't delete the folder.

oktett-8:
> @oktett-8 Can you update the code to remove the files under the assets instead of the whole asset folder? Our Frappe Operator https://github.com/vyogotech/frappe-operator does the same thing but doesn't delete the folder.

Well, I'm not the maintainer of `resources/core/main-entrypoint.sh`, but as far as I understand, the concept is to place a symbolic link to the container's part of the filesystem. So the mounted part of the filesystem (from docker volume to `/home/frappe/frappe-bench/sites`) symlinks the container part which updates with every release. Therefore (to allow creation of the symlink) the old asset directory has to be removed.

PR #1905, opened 2026-05-06
feat(images): toggle chromium installation

### LINKED ISSUE #1859: Unnecessarily Bloated v15 Docker Images since v15.98.1
## Description of the issue

Official production Docker images are unnecessarily larger since v15.98.1, due to an additional package added (chromium-headless-shell) for a feature in v16 (chromium pdf generator).

Check the official Docker image sizes for v15 here: https://hub.docker.com/r/frappe/erpnext/tags?name=v15&page=2

Observe that the image sizes are as follows:
- v15.98.1
   - linux/amd64 713.99 MB
   - linux/arm64 708.9 MB
- v15.98
   - linux/amd64 533.88 MB
   - linux/arm64 525.99 MB

This is a significant increase of 180 MB. This increase can be tracked down to this commit 8b3def7 in PR #1815 . There is an extra package installed `chromium-headless-shell` which contributes to this increase in size.

### Expected result

Since the feature is not available in v15, the package should not be installed for v15 builds.

My suggestion would be to install the package using a conditional for v16 builds only (call `apt-get install` a second time, but only for v16).

jslocomotor:
## Summary

Adds an optional `INSTALL_CHROMIUM` build argument to the bench, custom and production images.

Chromium installation remains enabled by default, but can now be disabled to reduce image size for setups that do not require Chromium-based PDF generation, i.e. Frappe Workbench 15 with ERPNext 15.

Closes #1859.

## Testing

* `docker build -f images/bench/Dockerfile images/bench`
* `docker build -f images/bench/Dockerfile --build-arg INSTALL_CHROMIUM=false images/bench`
* `docker build -f images/custom/Containerfile .`
* `docker build -f images/custom/Containerfile --build-arg INSTALL_CHROMIUM=false .`
* `docker build -f images/production/Containerfile .`
* `docker build -f images/production/Containerfile --build-arg INSTALL_CHROMIUM=false .`
* Verified `chromium-headless-shell` presence/absence accordingly
* `docker compose config`
* `pre-commit run --all-files`

  * `pyupgrade` failed locally on Python 3.14 with:
    `TypeError: cannot use a bytes pattern on a string-like object`
  * all other hooks passed

### DIFF
--- a/images/bench/Dockerfile
+++ b/images/bench/Dockerfile
@@ -75,7 +75,7 @@ RUN apt-get update \
     # For MIME type detection
     media-types \
     # Chromium
-    && if [ "$INSTALL_CHROMIUM" = "true" ]; then \
+    && if [ "$INSTALL_CHROMIUM" != "false" ]; then \
         DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
         chromium-headless-shell; \
     fi \
--- a/images/bench/Dockerfile
+++ b/images/bench/Dockerfile
@@ -4,6 +4,7 @@ LABEL author=frappé
 
 ARG GIT_REPO=https://github.com/frappe/bench.git
 ARG GIT_BRANCH=v5.x
+ARG INSTALL_CHROMIUM=true
 
 RUN apt-get update \
     && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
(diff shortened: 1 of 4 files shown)


images/bench/Dockerfile
```diff
@@ -75,6 +74,11 @@ RUN apt-get update \
     file \
     # For MIME type detection
     media-types \
+    # Chromium
+    && if [ "$INSTALL_CHROMIUM" = "true" ]; then \
```
DanielRadlAMR: i'd use `[ "$INSTALL_CHROMIUM" != "false" ]` instead

jslocomotor: Agreed, better handling of values.

jslocomotor:
Handling of values for INSTALL_CHROMIUM is now more permissive.

PR #1909, opened 2026-05-17
docs(contributing): add docker dev guide

harshith-ashok:
**Explain the details for making this change. What existing problem does the pull request solve?**

The existing development setup documentation assumes VS Code and the Dev Containers extension as the primary workflow. This leaves out developers who prefer working directly in the terminal, use a different editor, or simply want a lighter setup without IDE dependencies.

This PR adds a standalone guide for setting up a Frappe development environment using plain Docker and terminal commands, covering both x86 and ARM (Apple Silicon M1/M2/M3) systems. It consolidates the ARM64-specific notes from `docs/01-getting-started/03-arm64.md` into a single actionable walkthrough rather than scattered references.

The guide addresses several real gotchas that aren't documented anywhere in the current repo:

- Redis containers expose port `6379` internally regardless of host-side port mappings — using the wrong port in `common_site_config.json` causes `socketio` to silently fail
- The `sed` Procfile fix is required when Redis is running in separate containers, otherwise `bench start` crashes immediately
- MariaDB 11.8 (the current default image tag) may cause issues while creating a new site using `new-site` check (Issue #1908)
- Port conflicts on macOS (8000, 9000) are common due to system services and need explicit workarounds
- Assets must be built with `bench build` before first use or the UI loads completely unstyled

Edit: Removed `mariadb` version mismatch
Edit 2: Re added `mariadb` version notice

### DIFF
--- a/docs/05-development/04-alternate-setup.md
+++ b/docs/05-development/04-alternate-setup.md
@@ -6,6 +6,9 @@ title: Docker Development Setup
 
 A complete guide for setting up a Frappe development environment on x86 and ARM based computers running UNIX based OSes by running containers directly and working inside them via the terminal. No VS Code Dev Containers extension needed.
 
+> [!IMPORTANT]
+> Devcontainers are the intended development setup for Frappe Framework but in case you don't want to use that method follow these instructions to use the CLI directly instead
+
 ---
 
 ## Prerequisites
@@ -93,7 +96,7 @@ services:
 ```
 
 > Without this, Docker may pull amd64 images and emulate them via Rosetta — things will work but be noticeably slower.
-> Few users did note that the version of mariadb causes an error while creating a new site so in case you do encouter it change the version from `11.8` to `10.8`
+> Ensure mariadb version is set to `10.8` since Frappe Framework is not yet tested for the default version `11.8`
 
 ---
 
--- a/docs/05-development/04-alternate-setup.md
+++ b/docs/05-development/04-alternate-setup.md
@@ -96,7 +96,6 @@ services:
 ```
 
 > Without this, Docker may pull amd64 images and emulate them via Rosetta — things will work but be noticeably slower.
-> Ensure mariadb version is set to `10.8` since Frappe Framework is not yet tested for the default version `11.8`
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ The full `frappe_docker` documentation is available in [`docs/`](docs/) and publ
 - **Running in production:** [Production docs](docs/03-production/)
 - **Operating a deployment:** [Operations docs](docs/04-operations/)
 - **Development workflows:** [Development](docs/05-development/01-development.md)
+- **Docker Container Setup for Development** [Docker Setup](docs/10-development/01-setup.md)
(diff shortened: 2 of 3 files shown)


DanielRadlAMR:
There is already a development folder in the docs, so we should avoid creating an additional one.

I also don’t think this should be presented directly in the README, since we want to keep devcontainers as the standard development setup. This can instead be documented as an additional option for those who prefer not to use a devcontainer.

harshith-ashok:
Thanks for the feedback Daniel, How about I add it under `05-development` and call it `alternate-method.md`?

DanielRadlAMR:
You are very welcome, and thanks for the contribution in the first place!

Sounds good to me — maybe call it `alternate-setup.md`. Also, please state clearly somewhere that devcontainers are the intended development setup, and that this documentation is only meant for people who, for whatever reason, do not want to or cannot use devcontainers.

harshith-ashok:
Hi again Daniel, I have made the changes please let me know if there is anything else I need to change or add, And thanks again for taking the time to go through my PR 😊

docs/05-development/04-alternate-setup.md
```diff
@@ -0,0 +1,255 @@
...
+    # ...
+
+  redis-queue:
+    image: redis:6.2-alpine
+    platform: linux/arm64
+    # ...
+```
+
+> Without this, Docker may pull amd64 images and emulate them via Rosetta — things will work but be noticeably slower.
+> Ensure mariadb version is set to `10.8` since Frappe Framework is not yet tested for the default version `11.8`
```
DanielRadlAMR: Did you encounter any specific issues with MariaDB `v11.8`?

The setup should use the default versions. We’ve been using `v11.8` for quite some time and have never encountered any reproducible errors. Therefore, please remove this note.

DanielRadlAMR (changes requested):

harshith-ashok:
I have removed the line and made a commit please do check. This is like my first time trying to contribute so please excuse im just clickind random buttons 😂. And I apologise again for the lot of back and forth. Thanks

DanielRadlAMR:
Have you tried using `installer.py` instead of setting up Bench manually?

This could also be added to the documentation. Let me know whether you’re planning any further modifications. If so, I’ll wait before merging; otherwise, I’ll go ahead and merge it.

harshith-ashok:
I did try out the `installer.py` but that means installing bench locally and just pointing to the other services through docker which might not be the route people might want to take especially with server environments. 

Yeah, I dont have any more modifications from my side so you can go ahead and merge it Daniel. Thanks again.

DanielRadlAMR:
> I did try out the `installer.py` but that means installing bench locally and just pointing to the other services through docker which might not be the route people might want to take especially with server environments.

Hm, I’m not quite sure what you mean.

The script should do exactly the same as your steps 7, 8, 10, and 11.

harshith-ashok:
OK nevermind I think I misunderstood what you meant. Yes the `installer.py` is doing the same thing as the "manual" method. I faced some issues with the python version when i tested it using that method a while ago. 

I wasn't able to recreate it but the manual method eliminates the need for an extra dependencies and I did find it to be more consistent across all platforms

PR #1919, opened 2026-06-08
chore(devcontainer):add opinionated extensions

DanielRadlAMR:
* Remove `vscodeintellicode` (deprecated).
* Add Vue support, as it has been used throughout the framework since v16.
* Add the opinionated formatters `Prettier` and `Ruff`.


DanielRadlAMR:
@Rocket-Quack Since these extensions are somewhat opinionated, I'd appreciate your feedback on whether we should include them.

Rocket-Quack:
I like that because so we are moving a bit more in the direction of the recommended tools

PR #1953, opened 2026-08-04
fix(devcontainer): rely on VS Code port forwarding

DanielRadlAMR:
# Summary
Adapt the development setup to Frappe’s new localhost-only development server binding introduced by frappe/frappe#41516.
The devcontainer now relies on VS Code port forwarding instead of Docker-published ports.
# Changes
- Remove Docker port mappings for development ports.
- Always forward the primary ports:
    - 8000 — Frappe web
    - 9000 — Frappe realtime
    - 6787 — asset watcher

- Automatically forward additional Frappe services within:
    - 8000-8099 for web servers
    - 9000-9099 for realtime servers

- Ignore unrelated automatically detected ports.
- Explicitly configure VS Code’s port-forwarding behavior.
- Update CLI-only development documentation to explain how to:
    - set FRAPPE_BIND_ADDR=0.0.0.0;
    - publish the required ports through Docker;
    - restrict published ports to host loopback.

# Rationale
VS Code forwarded ports appear as localhost connections inside the container, so they continue to work when Frappe binds its development server to `127.0.0.1`.
Docker-published ports do not work with a service bound only to container loopback. Developers running the Compose setup without VS Code must therefore explicitly change the Frappe bind address and publish the required ports.
Production setups are unaffected because the production container runs Gunicorn with an explicit `0.0.0.0:8000` binding.

### DIFF
--- a/devcontainer-example/devcontainer.json
+++ b/devcontainer-example/devcontainer.json
@@ -1,6 +1,24 @@
 {
   "name": "Frappe Bench",
   "forwardPorts": [8000, 9000, 6787],
+  "portsAttributes": {
+    "8000-8099": {
+      "label": "Frappe web",
+      "protocol": "http",
+      "onAutoForward": "notify"
+    },
+    "9000-9099": {
+      "label": "Frappe realtime",
+      "onAutoForward": "silent"
+    },
+    "6787": {
+      "label": "Frappe asset watcher",
+      "onAutoForward": "silent"
+    }
+  },
+  "otherPortsAttributes": {
+    "onAutoForward": "ignore"
+  },
   "remoteUser": "frappe",
   "customizations": {
     "vscode": {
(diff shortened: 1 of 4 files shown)


Rocket-Quack (approved):
LGTM

DanielRadlAMR:
@Rocket-Quack, would you consider this a breaking change, since it might break development setups that don’t use VS Code?

Rocket-Quack:
@DanielRadlAMR yes these are breaking changes but as this only affects CLI Docker Compose setups I would still implement it for the benefit of everyone, as we can accept the migration because it is not that complicated

As long as this is mentioned in a release note or the documentation it should be fine as I am assuming most people will have carried out development using VS Code, because according to the documentation, that was the easyiest method.
And anyone still using a CLI environment can continue to use the migration path with `FRAPPE_BIND_ADDR=0.0.0.0`

So i think it would be fine🦆

PR #1958, opened 2026-08-13
fix(tests): grant MariaDB site access broadly

### LINKED ISSUE #1952: mariadb throws access denied because user has only one container ip access granted
## Description of the issue
`bench new-site` only grants access to one docker container (.5 is in my case the scheduler), so all other services gets an access denied.
```
SELECT User, Host FROM mysql.user WHERE User='_userId';
+-------------------+------------+
| User              | Host       |
+-------------------+------------+
| _userId | 172.26.0.5 |
+-------------------+------------+
```

`MySQLdb.OperationalError: (1045, "Access denied for user '_userId'@'172.26.0.6' (using password: YES)")`
I used the layered Containerfile and version-16. frappe, erpnext and hrms installed.

1. docker compose up -d
2. docker compose logs -f
3. ...

## Workaround
Grant privileges for the app user to % hosts:
```
  create-site:
    <<: *backend_image
    entrypoint:
      - bash
      - -c
    command:
      - |
        wait-for-it -t 120 db:3306
        wait-for-it -t 120 redis-cache:6379
        wait-for-it -t 120 redis-queue:6379
        if [ ! -f sites/${SITE_NAME}/site_config.json ]; then
          bench new-site \
            --admin-password=${ADMIN_PASSWORD} \
            --db-root-username=root \
            --db-root-password=${DB_ROOT_PASSWORD} \
            --install-app erpnext \
            --install-app hrms \
            --set-default \
          ${SITE_NAME};
        fi
        /home/frappe/frappe-bench/env/bin/python3 - <<'PYEOF'
        import json, MySQLdb
        site = "${SITE_NAME}"
        cfg = json.load(open(f"sites/{site}/site_config.json"))
        conn = MySQLdb.connect(host="db", user="root", passwd="${DB_ROOT_PASSWORD}")
        cur = conn.cursor()
        cur.execute(
            f"GRANT ALL PRIVILEGES ON `{cfg['db_name']}`.* TO '{cfg['db_user']}'@'%' "
            f"IDENTIFIED BY '{cfg['db_password']}'"
        )
        cur.execute("FLUSH PRIVILEGES")
        conn.commit()
        print(f"Grant auf '%' erweitert fuer {cfg['db_user']}")
        PYEOF
```

faizyyawan:

Fixes MariaDB site creation coverage so Docker test sites grant access from all container hosts, matching the documented `bench new-site --mariadb-user-host-login-scope=%` usage. This protects backend, scheduler, and queue containers from receiving a DB user limited to only the creating container IP.

Verification:
- `python3 -m pytest tests/test_site_creation.py -q`
- `python3 -m py_compile tests/conftest.py tests/test_site_creation.py`

Closes #1952

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -73,7 +73,7 @@ def frappe_site(compose: Compose):
     site_name = "tests.localhost"
     compose.bench(
         "new-site",
-        # TODO: change to --mariadb-user-host-login-scope=%
+        "--mariadb-user-host-login-scope=%",
         "--no-mariadb-socket",
         "--db-root-password=123",
         "--admin-password=admin",
@@ -97,7 +97,7 @@ def erpnext_site(compose: Compose):
     site_name = "test-erpnext-site.localhost"
     args = [
         "new-site",
-        # TODO: change to --mariadb-user-host-login-scope=%
+        "--mariadb-user-host-login-scope=%",
         "--no-mariadb-socket",
         "--db-root-password=123",
         "--admin-password=admin",
--- a/tests/test_site_creation.py
+++ b/tests/test_site_creation.py
@@ -1,0 +1,30 @@
+import pytest
+
+from tests import conftest
+
+
+@pytest.fixture(autouse=True, scope="session")
+def frappe_setup():
+    yield
+
+
+class FakeCompose:
+    def __init__(self):
+        self.bench_calls = []
+        self.compose_calls = []
+
+    def __call__(self, *cmd):
+        self.compose_calls.append(cmd)
+
+    def bench(self, *cmd):
+        self.bench_calls.append(cmd)
+
+
+@pytest.mark.parametrize("fixture", (conftest.frappe_site, conftest.erpnext_site))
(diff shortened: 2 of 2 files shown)


DanielRadlAMR (changes requested):
Thanks for picking this up. I think the TODO can be removed now, but we should replace `--no-mariadb-socket` rather than use both options. `--no-mariadb-socket` is a deprecated alias that already sets `mariadb_user_host_login_scope = "%"`, so the current patch doesn't change the effective behavior and continues to exercise the deprecated option.

Please replace `--no-mariadb-socket` with `--mariadb-user-host-login-scope=%` in both fixtures.

I also don't think the new unit test buys us much: it tests the fixture implementation via `__wrapped__` rather than DB connectivity. We already exercise connections from backend, scheduler and queue containers in the integration suite, which is the regression we care about here. I'd prefer relying on that test and getting the v15/v16 CI matrix green.

Finally, I don't think this PR by itself should Closes #1952, since it only changes test setup and pwd.yml already uses the wildcard host scope.

PR #1971, opened 2026-08-20
feat(actions): container security checks

DanielRadlAMR:
## Summary
Add container-focused validation and security checks for Dockerfiles, Compose overrides, and built images.
## Changes
- Add Hadolint to pre-commit and the existing lint workflow
  - Keep existing warnings visible but non-blocking
  - Fail on high-confidence rules covering root users, missing image tags, and latest tags
- Add a declarative Compose validation registry
  - Validate 16 supported Compose configurations
  - Require every overrides/compose.*.yaml file to appear in at least one configuration
  - Allow contributors to add validation by updating tests/compose-configs.json
- Add report-only Trivy scans for base, build, and erpnext images on AMD64 and ARM64
  - Show findings in PR workflow summaries
  - Retain complete SARIF reports as workflow artifacts
  - Upload findings to GitHub Code Scanning only on pushes to main
  - Do not fail builds during the initial rollout
- Add weekly Dependabot checks for proxy images
  - nginxproxy/nginx-proxy
  - nginxproxy/acme-companion
  - traefik
- Pin proxy images:
  - nginxproxy/nginx-proxy:1.11.6-alpine
  - nginxproxy/acme-companion:2.8.2
- Document the container validation workflow and contribution process
Database and Redis image updates remain excluded because their supported versions are defined by the Frappe framework.


DanielRadlAMR:
Trivy did find a substantial number of issues here. I'll check whether these come from Frappe dependencies that we can't directly influence, or whether we're missing something on our side.

DanielRadlAMR:
Essentially, I updated the Node version for v15 to 22, since Node 20 is already EOL. I also removed the fixed minor versions from our workflows (e.g. pointing to Python 3.14 instead of 3.14.2), so we automatically get the latest minor release with every run.

This already reduced the number of Trivy findings significantly, but there are still quite a few left.

Most of those are findings in dependencies, especially in the Frappe apps. I’m not quite sure how we want to proceed with these. The findings are very insightful, but there’s still a significant number of them, and I don’t want to bloat our code scanner results with issues we can’t directly influence.

At the same time, I don’t want to lose those insights either. They’re also a good starting point for updating dependencies in the Frappe apps themselves.

Rocket-Quack:
Sorry for the delay i have been really busy lately💻

It looks good 👍 
In the future we should have a plan for how to fine-tune the whole thing and maybe incorporate other quality gates since this currently serves only as a reporting tool so to speak.

But other than that LGTM 🫡

Rocket-Quack (approved):

DanielRadlAMR:
No worries, I didn't expect you to check it right away :)

Yeah, I deliberately introduced it just as a reporting tool for now. I want to gain some experience with it before turning it into an actual security gate. My concern is that many of the findings will come from dependencies that we cannot directly influence.

But I absolutely agree that adding some security-related gates is the right direction going forward.

PR #1987, opened 2026-09-07
feat(override): Add bind mount overrides for docker volumes

### LINKED ISSUE #1986: Support bind mount via environment variables

The current docker compose forces the use of named volume, making it not possible for a bind mount without modifying the compose file. It would be nice to have a few environment variables for configuring whether to use bind mount.

**Describe the solution you'd like**

A few of environment variables such as: 
- `SITES_DATA_LOCATION`
- `DB_DATA_LOCATION`
- `REDIT_DATA_LOCATION`
- etc

If the env variables are empty, it will fallback to the named volumes. As a side note, when bind mount is used, the named volume is still created although it's unused. This is quite minor since it's an empty unused volume. I think it's probably simpler to just let docker create the unused volume. But, it's actually possible to make docker to not create the volume using `external` attribute for the volume (see https://docs.docker.com/reference/compose-file/volumes/#external). 

**Describe alternatives you've considered**

A separate compose overrides `compose.bind-mount.yml` specifically for bind mount. But this is rather complicated because different overrides requires different volumes (e.g. redis and postgres overrides requires it's own volume). I think it would be simpler to let each overrides handles its own volume.

**Additional context**
N/A

lungsin:
Fixes #1986 

This PR does the following changes:
- Adds overrides for each docker volume so that it can be bind-mounted to the host directory. 
- Adds documentations for usage and caveats.

Currently, there's no way to do a bind mount beside customizing the compose file or adding a custom override. This PR adds some overrides for all volumes used (`sites`, `db-data`, and `redis-queue-data`).


DanielRadlAMR:
Could you take a look at https://github.com/frappe/frappe_docker/blob/main/docs/09-concepts/02-docker-bind-mounts.md?

Do we need to update these docs to reflect your changes?

lungsin:
Ah yes, we need to update it (didn't know that there's already a dedicated bind mount docs haha). I've updated the docs to reflect the changes. I'm omitting the erpnext specific setup since this docs seems to have generic educational content about docker bind mount.

DanielRadlAMR:
Yeah, this seems like fairly generic educational content. But I think we should somehow add a link to the overrides here. What’s your take?

Maybe just mention that there are overrides that make the bind-mount setup easier, briefly note the requirement that the path must already exist, and explain that they basically just define the volumes as you described above.

Or we don’t even need to go into that much detail — we could simply mention that there are overrides for this and link to `05-overrides`, which you already edited.

lungsin:
Ah I see! Yeah, I like the idea to simply mention the overrides for this and link to `05-overrides`. Also, the requirement that the path must already exist is mentioned on the example configuration (this was how [official docker](https://docs.docker.com/reference/compose-file/volumes/#driver_opts) documents it as well).

DanielRadlAMR:
@lungsin pls rebase. The failing tests are not due to your changes.

lungsin:
@DanielRadlAMR Done!

PR #1991, opened 2026-09-09
fix: Update Containerfile for setup Ansible

HUMENTH:

Mandatory requirement of Ansible for "Suite Cloud" app for self-hosted instances.

[ansible must be available on the bench host](https://github.com/frappe/suite_cloud/blob/develop/README.md#:~:text=ansible%20must%20be%20available%20on%20the%20bench%20host)

### DIFF
--- a/images/layered/Containerfile
+++ b/images/layered/Containerfile
@@ -15,7 +15,7 @@ RUN --mount=type=secret,id=apps_json,target=/opt/frappe/apps.json,uid=1000,gid=1
   if [ -f /opt/frappe/apps.json ] && [ -s /opt/frappe/apps.json ]; then \
     export APP_INSTALL_ARGS="--apps_path=/opt/frappe/apps.json"; \
   fi && \
-  bench init ${APP_INSTALL_ARGS}\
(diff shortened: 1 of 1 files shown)


DanielRadlAMR:
I don't quite get what this is used for.

It sounds like something that belongs in a custom app setup rather than being generally available in the images.

HUMENTH:
> I don't quite get what this is used for.

@DanielRadlAMR ansible is a mandatory requirement for official **Frappe Mail** app as per current documentation. So, after few months of the stable release of Frappe Mail app, mostly ERPNext members will use this app _(my personal opinion)_. That's why I tried to update this requirement in base file of Frappe Docker.

DanielRadlAMR:
But Frappe Mail is the mail server itself, right?

Personally, I wouldn’t run my mail server in the same environment as my ERP system. I think those should be separated and connected via an API.

Therefore, I don’t see why we would want to add mail server setup requirements to all layered images. This seems app-specific to me.

HUMENTH:
Agreed with you.

PR #1994, opened 2026-09-18
feat: make the internal nginx server's listen port configurable

sujaldev:

I have added a `NGINX_LISTEN_PORT` environment variable that can be used to configure what port the internal nginx server will listen on.

(This is intentionally not named something along the lines of `NGINX_LISTEN` as it is also used in some contexts where only a port value is acceptable.)

While this might be end up being useful for various other scenarios, my [personal use case](https://github.com/sujaldev/infra) is that I'd like to configure the internal nginx server inside the frontend container to work with systemd socket activation and a unix socket.

This currently requires changing the hard-coded `listen 8080` directive in `resources/core/nginx/nginx-template.conf` to `listen unix:/path/to/my/socket`. If this PR is merged, I can simply configure `NGINX_LISTEN_PORT=unix:/path/to/my/socket` instead of maintaining a patch.


DanielRadlAMR:
Overall, I think this is a useful addition. Making the internal nginx listen port configurable removes the hard-coded `8080` assumption and gives custom deployments more flexibility.

One thing I’d suggest is documenting the Unix socket use case separately from the normal environment-variable docs.

`NGINX_LISTEN_PORT` is substituted directly into nginx’s `listen` directive, so values like `unix:/path/to/socket` can work as well. That’s an interesting advanced use case, but it may break overrides or reverse proxies that expect a normal TCP port.

I’d keep the env-var docs focused on custom TCP ports and add a short note in a more advanced section explaining that Unix sockets are possible, but should only be used if you understand the surrounding setup and compatibility implications.

Apart from that, the change looks good to me.

docs/02-setup/04-env-variables.md
sujaldev, line 178: How about a tip right here?


> Tip: `NGINX_LISTEN_PORT` is substituted directly into the internal Nginx's [listen](https://nginx.org/en/docs/http/ngx_http_core_module.html#listen) directive, so values like `unix:/path/to/socket` can also be used. However, this is an advanced use case that may break some compose overrides that strictly expect a port value.

DanielRadlAMR: I would keep this doc focused on the TCP nginx port setup. That should always work, the naming fits, and it’s the “expected” setup.

I’d treat the Unix socket setup more like a “known workaround” (or whatever we want to call it). It should really only be used if you know what you’re doing and explicitly want this kind of setup.

Therefore, I’d also document it somewhere else. I’m a bit worried that putting it in such a prominent position might encourage people who aren’t fully aware of the implications to try it anyway 😄 and then we’ll end up with tickets because their setup doesn’t work.

sujaldev: Where do you suggest we put this? I don't see an existing place in the docs that'll be a good fit, and creating a new page just for this seems overkill.

Also, are we sure we even want this documented? If someone is deliberately replacing the listen directive with a unix socket, they're either passing said socket as an existing file descriptor, or they're mounting it as a volume. Either way, they're going beyond what the default compose files provide, and will need to understand how the containers are wired together.

Since documenting the latter part is beyond the scope of these docs, I don't think it's unreasonable to expect users to put two and two together here.

DanielRadlAMR: From what I’m seeing, some setups are currently explained in `03-production`, although I have to say the name might be a bit misleading.

I mean, we don’t *have to* document it. I just thought that if it’s useful to you, it might be useful to others as well, and therefore could be documented somewhere. All of these additional considerations could be mentioned there too.

If it gets very technical, it could also live in `08-reference`. References are generally more about broader technical background rather than one specific setup, though. So if we put it there, we could perhaps explain the port more generally—what it does, why it’s needed, and how it works—and then mention that it can be switched to a Unix socket, together with all the relevant considerations.

sujaldev:
I've added it as a new document under `03-production`. What do you think? Could be used to document other such setups too in the future.

DanielRadlAMR:
seems fine to me

But the pipeline needs to be green

PR #1995, opened 2026-09-19
feat(images): upgrade to debian trixie and support openshift arbitrary uid

revant:
### Summary

- **OpenShift `restricted-v2` & Arbitrary Non-Root UID Compliance**:
  - Sets default user `frappe` to GID `0` (`-g 0`).
  - Sets group ownership (`chgrp -R 0`) and group write permissions (`chmod -R g=u`) on `/home/frappe`, `/etc/nginx`, `/var/log/nginx`, `/var/lib/nginx`, `/run`, and `/etc/passwd`.
  - Adds `umask 0002` in entrypoints so files/logs created at runtime remain group-accessible on shared/mounted volumes.
  - Adds dynamic `/etc/passwd` entry generation in `main-entrypoint.sh` for arbitrary dynamic UIDs without user database lookups failing.
  - Adds integration test `test_arbitrary_uid_execution` running with `--user 1000680000:0`.

- **Debian Trixie Upgrade**:
  - Upgrades default `DEBIAN_BASE` to `trixie` across production, custom, and bench images.
  - Replaces deprecated `libtiff5-dev` with `libtiff-dev`.
  - Maintains `wkhtmltopdf` (with patched Qt) compatibility using the bookworm release package.
  - Updates `CONTRIBUTING.md` maintenance instructions.

- **Buildx Attestations & OCI Labels**:
  - Enables native Buildx attestations for SPDX SBOM and SLSA provenance (`type=sbom`, `type=provenance,mode=max`) in `docker-bake.hcl`.
  - Adds standard OCI source and URL labels.


DanielRadlAMR (commented):
A few things I noticed while going through this:

* I’m not quite sure about the `/etc/passwd` handling here. As far as I understand it, the image already creates `frappe` with UID `1000`, and then the entrypoint can append another `frappe` entry with the arbitrary runtime UID if `whoami` fails. Together with making `/etc/passwd` group-writable, this seems like we could end up with two `frappe` entries with different UIDs, while also allowing the container user to modify `/etc/passwd`. Maybe I’m missing the reason this is necessary, especially on OpenShift/CRI-O. Did you consider whether this could cause confusing user lookups or whether there is a way to avoid making `/etc/passwd` writable?

* Small style thing: do we need the `\` at the end of all the comment lines in the Dockerfiles? For example:

  ```dockerfile
  # For backups \
  ```

  AFAIK comments don’t need line continuation, so to me this mostly adds visual noise. I’d personally leave the backslashes only on the actual command/package lines.

* The new arbitrary UID test looks useful. A few possible improvements I noticed:

  * explicitly assert that the UID is actually `1000680000`, not only that the GID is `0`
  * the docstring says `/tmp` is tested, but the test currently doesn’t write there
  * since the PR also changes the nginx permissions/entrypoint, it might be useful to test the `frontend` container with an arbitrary UID as well, e.g. at least `nginx -t` or a successful startup

* Did you also check whether the docs need to be updated for these changes? A few places seem stale after this PR:

  * `docs/02-setup/04-env-variables.md` still describes the `sites` owner as UID/GID `1000` and uses `chown 1000:1000`
  * `overrides/compose.bind-mount-sites.yaml` also says `uid/gid 1000` / `chown -R 1000:1000`
  * `example.env` says the sites directory is owned by the UID/GID the service runs as, with `1000` for sites
  * `docs/02-setup/02-build-setup.md` still says `DEBIAN_BASE` defaults to `bookworm`, and describes it as “bench only”, while the default is now `trixie` and the production/custom builds use that argument too

  Especially with the new arbitrary-UID/OpenShift support, it might also be worth documenting the intended ownership/permission model for mounted `sites` directories (arbitrary UID + GID 0 / group-writable).

DanielRadlAMR (commented):
A few small things I noticed:

* There are still a couple of comments with trailing `\` in `images/bench/Dockerfile`.
* The test docstring now says `"Frontend nginx configuration test and startup"`, but as far as I can tell it only runs `nginx -t`, so it tests the configuration but not an actual nginx startup.
* `/etc/passwd` is still made group-writable. Is that intentional / still required for the arbitrary UID handling?
* The bench Dockerfile still hardcodes `debian:trixie-slim`, and `DEBIAN_BASE` is not passed to the bench target either. Is that intentional, or should the bench image use the same `DEBIAN_BASE` argument as the other images?

DanielRadlAMR (approved):
LGTM!

I assume you tested frappe and erpnext v16 to run on trixie?

revant:
ERPNext v16 works with trixie.

[pwd-trixie.webm](https://github.com/user-attachments/assets/bf07364f-467e-4a05-be47-c54894dda44b)

